PR #2, opened 2022-12-12
Add git ignores for build artefacts

pkendall64:
When building the application the file tree is left with quite a lot of files that should really be ignored, as I'm sure they should not be committed to the git repository.


.gitignore
```diff
@@ -0,0 +1,8 @@
+build
+linux-4.9
+media
+toolchain
+mkapp/image
+mkapp/ota_app
```
bkleiner, line 6: we probably want to track (and delete from the repository) `mkapp/app/app/HDZGOGGLE` here too

bkleiner: seems the binary got modified, not removed.

pkendall64: Yes, fixed and force pushed

PR #4, opened 2022-12-12
Readme

pkendall64:
Two commits to in this PR.
First just fixes a few small errors in the original readme, second adds a section on building the image in a docker container so it can be built on any system that supports docker (with volume mounts).

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,14 +12,14 @@
 ## 2. Setup cmake environment
 
 ```
-~/hdz_goggle/workspace/code$./setup.sh
+~/hdz_goggle/code$ ./setup.sh
 ```
 
 # Build code
 
 ```
- ~/hdz_goggle/code/$cd build
- ~/hdz_goggle/code/build$make clean all
+ ~/hdz_goggle/code/$ cd build
+ ~/hdz_goggle/code/build$ make clean all
 ```
 
 the firmware is generated on ~/hdz_goggle/code/out
--- a/README.md
+++ b/README.md
@@ -23,3 +23,36 @@
 ```
 
 the firmware is generated on ~/hdz_goggle/code/out
+
+
+# Building in a container
+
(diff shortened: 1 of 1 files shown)


razzeee:
Might be neat to handle this in a devcontainer https://code.visualstudio.com/docs/devcontainers/create-dev-container

bkleiner:
> Might be neat to handle this in a devcontainer https://code.visualstudio.com/docs/devcontainers/create-dev-container

That seems like a neat solution indeed. It has a spec and everything!
Can somebody comment on how much of a pain docker is on Windows these days?
Last time (a while ago) i had to set it up there it gave me all kinds of grief.
Volume mounts seemed especially problematic.

ligenxxxx:
@pkendall64 I have not used docker, can you confirm that these modifications are correct?

pkendall64:
> @pkendall64 I have not used docker, can you confirm that these modifications are correct?

Yes, I used the instructions in the PR to actually do a build of the firmware on a MacOS system.

ligenxxxx:
I want to start from the step of git clone. It only needs some simple operations, so that rookies can also successfully build firmware.
- ~/hdz_goggle/code$ is a wrong path, there is no folder named code.
- ./setup.sh needs to modify the permissions before it can run.

ligenxxxx:
> > @pkendall64 I have not used docker, can you confirm that these modifications are correct?
> 
> Yes, I used the instructions in the PR to actually do a build of the firmware on a MacOS system.

I will test it on windows.

razzeee:
> I want to start from the step of git clone. It only needs some simple operations, so that rookies can also successfully build firmware.


https://github.com/hd-zero/hdzero-goggle/pull/8 should be way simpler for rookies, at least if we assume vscode is used

pkendall64:
> can we close this one in favor of #8?

Absolutely we can 😄

PR #5, opened 2022-12-12
update toolchain

bkleiner:
the gcc version in the toolchain as shipped is ancient. (2016!)
changes the setup script to fetch a more up-to-date version from http://musl.cc/
and streamlines other portions of the process.
marking this as a draft, as i can not verify operation right now. (goggle is supposed to arrive in a couple days)


bkleiner:
closing in favor of a devcontainer based approach

PR #7, opened 2022-12-12
`out` should be a directory

pkendall64:
After reading through the codebase and looking at the cmake files, it looks like the the final artefact `out` should actually be a directory that contains the `HDZGOGGLE` binary and the `HDZERO_GOGGLE-x.x.x.bin` firmware file that should be copied to the root directory if the SD card on the goggles for installation/upgrade.


CMakeLists.txt
```diff
@@ -186,6 +186,7 @@ target_link_directories(${PROJECT_NAME} PRIVATE
 
 add_custom_command(TARGET ${PROJECT_NAME} POST_BUILD 
 											COMMAND size -A ${PROJECT_NAME}
+											COMMAND mkdir -p ${PROJECT_SOURCE_DIR}/out
 											COMMAND cp ${PROJECT_NAME} ${PROJECT_SOURCE_DIR}/out
 											COMMAND cp ${PROJECT_NAME} ${OS_APP_PATH}
 											COMMAND cd ${OS_APP_PATH} && cd ../../ && rm -f ./ota_app/hdzgoggle_app_ota*.tar && ./mkapp_ota && ./make_fw.sh
```
cruwaller, line 192: Could it be a good idea to change `cd ${OS_APP_PATH} && cd ../../` to `cd ${PROJECT_SOURCE_DIR}/mkapp/`

cruwaller: ...and add `/` in end of the move command below? `${PROJECT_SOURCE_DIR}/out` -> `${PROJECT_SOURCE_DIR}/out/`

pkendall64: Changes made as suggested 👍

cruwaller (approved):
Thanks PK!

PR #8, opened 2022-12-12
add devcontainer setup

bkleiner:
- add support for the [vscode devcontainer integration](https://code.visualstudio.com/docs/devcontainers/containers)
- extend `setup.sh` to take care of the bootstrap process
- update readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,26 +1,33 @@
-# Initial setup
+# HDZero Goggle Firmware
 
-## 1. Extract supporting libraries and head files
+## Environment Setup
 
-```
-~/hdz_goggle/code$ tar xzvf linux-4.9.tar.gz
-~/hdz_goggle/code$ cat media.tar.gzaa media.tar.gzab > media.tar.gz
-~/hdz_goggle/code$ tar xzvf media.tar.gz
-~/hdz_goggle/code$ tar xzvf toolchain.tar.gz
-```
+The firmware can either be built in a [devcontainer](https://containers.dev/) or natively on a linux machine.
 
-## 2. Setup cmake environment
+### Devcontainer Setup
 
-```
-~/hdz_goggle/workspace/code$./setup.sh
-```
+This repository supports the [vscode devcontainer](https://code.visualstudio.com/docs/devcontainers/containers) integration.  
+To get started install docker, vscode and the decontainer extension.  
+A [prompt](https://code.visualstudio.com/docs/devcontainers/create-dev-container#_add-configuration-files-to-a-repository) to reopen this repository in a container should appear.
 
-# Build code
+### Native Setup
+
+CMake is required for generating the build files.  
+A bash script is supplied to take care of the bootstrap process:
 
 ```
- ~/hdz_goggle/code/$cd build
- ~/hdz_goggle/code/build$make clean all
+~/hdzero-goggle$ ./setup.sh
 ```
 
-the firmware is generated on ~/hdz_goggle/code/out/HDZERO_GOGGLE-x.x.x.bin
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -1,25 +1,22 @@
...
-~/hdz_goggle/code$ cat media.tar.gzaa media.tar.gzab > media.tar.gz
-~/hdz_goggle/code$ tar xzvf media.tar.gz
-~/hdz_goggle/code$ tar xzvf toolchain.tar.gz
-```
+This repository supports the [vscode devcontainer](https://code.visualstudio.com/docs/devcontainers/containers) integration.  
+To get started install docker, vscode and the decontainer extension.  
+A [prompt](https://code.visualstudio.com/docs/devcontainers/create-dev-container#_add-configuration-files-to-a-repository) to reopen this repository in a container should appear.
 
-## 2. Setup cmake environment
+## Building locally
```
razzeee, line 9: I find this a bit confusing, as this is only a setup step and not actually building anything, so making that clear would be helpful. So the old naming was probably better here.

bkleiner: ...i hear you, what do you suggest? how about "Setup local environment"? i want to make sure that it is conwayed that is a either or deal.

razzeee:
Building seems to fail for me with `/bin/sh: 1: ./mkapp_ota: Permission denied` which might be fixed by one of the other PRs setting executeable flags?

It also seems to complain a lot about 
```
arm-openwrt-linux-muslgnueabi-gcc.bin: warning: environment variable 'STAGING_DIR' not defined
```

bkleiner:
> Building seems to fail for me with `/bin/sh: 1: ./mkapp_ota: Permission denied` which might be fixed by one of the other PRs setting executeable flags?
yup https://github.com/hd-zero/hdzero-goggle/pull/3

> It also seems to complain a lot about
> 
> ```
> arm-openwrt-linux-muslgnueabi-gcc.bin: warning: environment variable 'STAGING_DIR' not defined
> ```
yup...that toolchain is not exactly optimal (warning is caused by a wrapper intended for the openwrt buildroot), but baby steps :)

razzeee:
New readme is lovely.

Very nitpick, me problem,
I don't get why people do this cmd prefixing:

```
~/hdzero-goggle$ 
```

It only breaks copy and paste, is that what it's for? :)

bkleiner:
> New readme is lovely.
> 
> Very nitpick, me problem, I don't get why people do this cmd prefixing:
> 
> ```
> ~/hdzero-goggle$ 
> ```
> 
> It only breaks copy and paste, is that what it's for? :)

yeah i am not a huge fan either, but i guess the idea is to make directory changes more visible.
every line explicitly shows you what CWD is expected.

ligenxxxx:
I modify setup.sh file format from dos to unix, then setup.sh can run in windows docker.

razzeee:
If it runs in docker, the unix format should work, as were already in the Debian container.

You probably meant for the local setup?

bkleiner:
> If it runs in docker, the unix format should work, as were already in the Debian container.
> 
> You probably meant for the local setup?

the issue seems to be that windows still defaults to checking out the repository with CRLF.
this then causes issues when the scripts are mounted into the container. 
so i introduced the gitattributes file, trying to force the it to LF, but i guess it doesn't change the line-endings on existing checkouts.

ligenxxxx:
Before we merge this PR, we should also test it on the ubuntu system of the virtual machine.

razzeee:
Just tried again and it seems to build fine for me on Manjaro Linux, using the devcontainer.

PR #14, opened 2022-12-23
ELRS backpack integration

pkendall64:
This is a cleanup and of the ESP32 flashing code, building the flashing into the goggle code rather than a separate executable, with progress of each binary being flashed to the ESP32.

The ELRS backpack functionality currently supported by this PR are
1. Channel changing/setting from VTx Admin menu in ELRS Lua.
2. DVR start/stop (with configurable delays) assigned to an Aux channel/switch

To be done, is the OSD support that ELRS backpack can provide (i.e. for putting race/lap times on screen).

To use this PR, you should also build the ELRS backpack for the HDZero VRx using this PR https://github.com/ExpressLRS/Backpack/pull/73


bkleiner:
Seems to work beautifully, tested:
- [x] flashing via the goggles
- [x] binding to the tx module (maybe a note to update the tx backpack to latested somewhere would be good)
- [x] channel switch (hdzero)
- [X] dvr toggle via AUX

Only thing i found counterintuitive, i would have expected the flash button to be located in the "Firmware" menu vs the "Connections" menu, but that might just be me.

pkendall64:
> Only thing i found counterintuitive, i would have expected the flash button to be located in the "Firmware" menu vs the "Connections" menu, but that might just be me.

I do agree, but that was just where it was initially. We can easily move it, if we get consensus.

Zuldan:
> Only thing i found counterintuitive, i would have expected the flash button to be located in the "Firmware" menu vs the "Connections" menu, but that might just be me.

Agreed. Better to move it to Firmware

PR #15, opened 2022-12-23
cmake improvments

bkleiner:
- adds [ninja build](https://ninja-build.org/) as an faster alternative to make to the devcontainer
- pass cmake configuration to vscode to allow its integration to produce functioning binaries
- move linux headers, softwinner, minIni and lvgl out of tree and compile them as cmake libraries
- drop now unneeded linux header tar
- merge `media` into the main application build
- adjust lvgl sidebar width, seems there was a discrepancy between the code as supplied and the pre-built library


pkendall64:
This is really good to get the code cleaned up a bit and into a more sensible layout.
I've create a PR against you branch doing the same thing with the esp-loader code.

PR #21, opened 2022-12-29
Refactor Log Library

bkleiner:
Replaces the existing liblog.so with our own implementation and uses it throughout the application.
Very simple right now, but i think its important that we establish a api so the log format can be unified.
File-logging could be easily added at a later date.


bkleiner:
Thanks for the review, appreciated! I am not sure how happy I am with this overall. It does tidy up the output quite a bit…

pkendall64:
> Thanks for the review, appreciated! I am not sure how happy I am with this overall. It does tidy up the output quite a bit…

Personally, I think it's a good idea to have a consistent logging framework across the whole application.

PR #22, opened 2022-12-30
import version 7.67.122.6

bkleiner:
import version 7.67.122.6 from divimath:

- import SNEAKY_FPV 's new font
- HDMI in 1080p60
- Less fan noise
- Show connected if RapidFire is on
- Analog adjusted slightly brighter
- Disabled color kill


bkleiner:
kind off, more than half of the divimath staff is currently sick, so there are not enough resources to fully transition everybody to this new git repo (they used svn previously). There are issues and feature request that need the attention of the few remaining divimath engineers however, so we are making due with manual syncs for the time being.

PR #25, opened 2023-01-01
Change autoscan to on/last/off

### LINKED ISSUE #23: Goggle does not Auto Scan  on boot up (firmware 2212022)
Existing Behaviour
=============
With Auto Scan setting set to
    Auto Scan = On
    Default      = HDZero

When powering up the goggles, the user is taken to the Scan menu.
In order to scan for a transmitting video signal from an aircraft, the user is forced to trigger a scan by pressing the left centre button.

Expected Behaviour
=============
When powering up the goggles with Auto Scan = On, the goggles should automatically scan, and give the user a choice of channels which were detected on the scan. Apart from removing redundant user input, which harms the user experience, this behaviour is consistent with the way that HDZero VRx products operate.

If Auto Scan = Off, and Default is set to HDZero, the goggles should automatically attempt to tune to the last HDZero channel which was selected on a previous session.

pkendall64:
Changes the "Auto Scan" option to `On`/`Last`/`Off`

- `On` does the auto scan
- `Last` tunes to last channel
- `Off` just displays the main menu on the scan page

Fixes #23

### DIFF
--- a/src/page/page_common.h
+++ b/src/page/page_common.h
@@ -65,10 +65,24 @@ typedef struct {
 	int channel;
 } setting_scan_t;
 
+typedef enum {
+    SETTING_AUTOSCAN_SCAN = 0,
+    SETTING_AUTOSCAN_LAST = 1,
+    SETTING_AUTOSCAN_MENU = 2
+} setting_status_t;
+
+typedef enum {
+    SETTING_SOURCE_LAST = 0,
+    SETTING_SOURCE_HDZERO = 1,
+    SETTING_SOURCE_EXPANSION = 2,
+    SETTING_SOURCE_AV_IN = 3,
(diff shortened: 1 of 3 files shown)


src/page/page_common.h
```diff
@@ -66,9 +66,9 @@ typedef struct {
 } setting_scan_t;
 
 typedef struct {
-	bool status;
-	int  last_source;
-	int  source; //0=Last mem, 1=HDZero,2= Expansion, 3=AV in,4=HDMI in
+	int status;	// 0=enable, 1=last, 2=disable
```
bkleiner, line 69: Would be neat if we started converting these to enums, juggling raw ints is gonna get confusing.

pkendall64: Good idea, I'll do that as part of this PR then.

PR #29, opened 2023-01-03
Backpack cleanup

pkendall64:
Moves the backpack flashing to the "Firmware" menu, as previously suggested.
Also adds colors to "Success", "FAILED" and "Timeout" status values.
Button text returns back to _default_ value when cursor is moved.


bkleiner:
Sorry, only got back to my workstation today, had a bit of backlog there. Could we get a rebase on this one?

pkendall64:
> Sorry, only got back to my workstation today, had a bit of backlog there. Could we get a rebase on this one?

Rebased and tested.

PR #32, opened 2023-01-03
Sub-menu mods

pkendall64:
Allows the pages to define the items that are selectable, so on the connections page the SSID and Broadcast items are not selectable as I believe the intent is that the "configure" option should bring up a "page" or "dialog" for configuring the wifi adapter settings.

Also, moves the `page_pack_t` instances into the matching `page_xxx.c` file.


bkleiner:
just a thought, should we rename the `page` folder to `ui` and move all the ui related files from `core` there?

pkendall64:
I think moving all the UI stuff to a `ui` would be a good idea as well.

w.r.t the Wifi configuration, theres also a broadcast flag, which applies to AP mode as to whether or not to broadcast the SSID. I have no idea if that is possible as I don't actually know the features of the wifi module 😄 

Maybe we just do the moving of the code and leave the other parts until the wifi module is available.

bkleiner:
> w.r.t the Wifi configuration, theres also a broadcast flag, which applies to AP mode as to whether or not to broadcast the SSID. I have no idea if that is possible as I don't actually know the features of the wifi module smile
> 
> Maybe we just do the moving of the code and leave the other parts until the wifi module is available.

Yeah let's wait until we have a better picture on what the module is capable of.

pkendall64:
Ok, I'll close this PR and think about moving all the UI related code into a `ui` folder.

PR #39, opened 2023-01-05
UI cleanup

bkleiner:


bkleiner:
tacked on some more changes tiding the player code and fixing #54

PR #53, opened 2023-01-05
Cleanup strings

razzeee:
These are various improvements to the strings. Keep in mind, that this isn't tested as I haven't gotten my googles yet.

Especially the statusbar change might cause problems, if it's too long now, but hopefully it's not, that's a place where it's really bugging me.

The correct handling of the back button, would probably also be a custom item, with a button pointing back. Basically similar to how we do the selection item. But this is still an improvement for now.


bkleiner:
Statusbar seems fine.

PR #59, opened 2023-01-06
page pack refactor

bkleiner:
this pr reworks the page pack interface to allow specifying handler functions for different page interactions,
this moves the complexity out of the main menu file and into the page files where the rest of logic usually resides.
this allows for generic handling of all pages in main_menu, greatly simplifying the process of adding new pages.

additionally:
- formats all page files 
- tidies some type definitions

for review i recommend skipping the first commit, it does not change any logic, but only formats the file according to the clang-format file

### DIFF
--- a/src/ui/page_autoscan.c
+++ b/src/ui/page_autoscan.c
@@ -1,27 +1,25 @@
 #include "page_autoscan.h"
 
+#include "minIni.h"
 #include "page_common.h"
 #include "ui/ui_style.h"
-#include "minIni.h"
 
-
-static lv_coord_t col_dsc[] = {100,150,180,220,180,160, LV_GRID_TEMPLATE_LAST};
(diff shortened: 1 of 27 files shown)


src/ui/page_autoscan.c
```diff
@@ -1,84 +1,92 @@
...
+    }
 }
+
+page_pack_t pp_autoscan = {
+    .p_arr = {
+        .cur = 0,
+        .max = 4,
+    },
+
+    .create = &page_autoscan_create,
```
pkendall64, line 87: You don't need to pass the address of the function as giving just the name provides the pointer to the function already.
i.e. without the `&`
```
.create = page_autoscan_create,
```

pkendall64 (commented):
Looks good.

Issue #37, opened 2023-01-04
Focus chart screen to aid IPD/focus adjustment


MBoerlage:
Request to add an option to display a full screen "**Back Focusing Chart**" image (black & white) to help with manual IPD/Focus adjustment. 

1. Select "Focus Chart" in goggle menu (to be added)
2. Display full screen image of "Back Focusing Chart" - its a black and white chart wheel in the center and small ones in each corner
3. Allow user to adjust mechanical IPD/Focus nobs on goggle
4. Use a button to return to the goggle menu, return to menu

The image should be stored in firmware (not on SD card), several "Back focus chart" examples can be found on goggle. Since its sole purpose is focus, it is preferred over colored TV test screens. It is crucial the image has small circles on each corner and one in the middle. 

Example applications are ORQA Pilot.One goggles, they have this as an option "Deep focus" which is awesome. Some users adjust IPD/focus looking at the camera footage (inaccurate) or their OSD on their camera footage (depending on where your OSD is, inaccurate). A back focusing chart will make sure you get the best sharpness out of your setup.

riveraj082:
This is a good idea.

riveraj082:
This is a good idea.

MBoerlage:
![image](https://user-images.githubusercontent.com/29292730/210698884-70aedc7f-32e5-4388-abef-cbfb474ab6e8.png)

EastCoaster:
Excellent idea, maybe even include the steps needed to dial in focus and IPD.

MBoerlage:
> i have never heard of these before, but does indeed seem like a useful addition. do you think you could provide a 1920x1080 version of this chart? i assume we would want it to fill the whole screen.

Please find attached - with many thanks and credits to Ian from the Houston FPV race community, he made one from scratch. Feel free to suggest any changes/adds. 

![Ian_Back_Focus_1080](https://user-images.githubusercontent.com/29292730/211003936-9e74c9ba-c517-4ec3-bbee-dec79c4fabe9.png)

bkleiner:
> > i have never heard of these before, but does indeed seem like a useful addition. do you think you could provide a 1920x1080 version of this chart? i assume we would want it to fill the whole screen.
> 
> Please find attached - with many thanks and credits to Ian from the Houston FPV race community, he made one from scratch. Feel free to suggest any changes/adds.
> 
[...]

PR #60, opened 2023-01-06
Feature focus chart

bkleiner:
adds a focus chart screen to the firmware as proposed by https://github.com/hd-zero/hdzero-goggle/issues/37

builds on #59


pkendall64 (commented):
Looks good to me. Flashed and tested 👍 
If we add more utility style things like this then it might be good to have Utilities menu with Focus Chart and any others in it. But that should wait till theres a need.

bkleiner:
> Looks good to me. Flashed and tested +1 If we add more utility style things like this then it might be good to have Utilities menu with Focus Chart and any others in it. But that should wait till theres a need.

yes i agree, however multiple levels of menus/screens will be much neater and easier to implement once we refactored more of the input & app state system, so this seemed like a good compromise.

PR #73, opened 2023-01-11
Disable horizontal chroma fir filter for vdpo

ligenxxxx:
The default FIR filter causes the OSD's chroma border to spread, eventually causing the OSD border to display the wrong color.
`dispw` will overwrite `0x06542018` to `0x00000022`, and `aww` needs to be run after `dispw` is executed.

### DIFF
--- a/src/driver/hardware.c
+++ b/src/driver/hardware.c
@@ -80,6 +80,7 @@ void Display_UI_init()
     I2C_Write(ADDR_FPGA, 0x84, 0x11);
 
     OLED_SetTMG(0);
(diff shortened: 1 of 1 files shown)


src/driver/hardware.c
```diff
@@ -80,6 +80,7 @@ void Display_UI_init()
     I2C_Write(ADDR_FPGA, 0x84, 0x11);
 
     OLED_SetTMG(0);
+	system("aww 0x06542018 0x00000044"); //disable horizontal chroma FIR filter.
```
bkleiner, line 83: can you fix the indentation here?

PR #78, opened 2023-01-13
add delay to sensor module loads

bkleiner:
the goggle application relies on the mcp3021 being registered _first_ so it gets assigned `device0` in the dev fs.
insmod is async, so when both load commands are sent back to back there is a good chance nct75 is assigned `device0`.
this PR adds a flat delay, mitigating the biggest risk. a better solution would be to have the application search the `name` files in the dev-fs.


bkleiner:
does not fix the underlying problem, investigating further....

PR #86, opened 2023-01-13
Update the spaces and styles to be more coherent within the file

AlexandreBonneau:
I'm not sure which style is used through the project, however I noticed some inconsistencies in that single main.c file.
This PR fixes the mixing of spaces with tabs, and makes the style coherent overall, whatever it may be.

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -43,34 +43,34 @@ static void load_ini_setting(void)
 	
 	FILE* fp;
 	fp=fopen("/mnt/UDISK/setting.ini","r");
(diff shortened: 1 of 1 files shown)


bkleiner:
the style is specified in the the [clang format file](https://github.com/hd-zero/hdzero-goggle/blob/main/.clang-format) inherited from the vtx project.
large, formatting, commits can inhibit being able to merge cleanly, so i have been trying to be strategic about _when_ to apply it.
now wouldn't be the worst time, but it should follow the clang-format specification.

Issue #27, opened 2023-01-02
Double tap menu button to quick cycle through input sources of HDZ/HDMI/Expansion Module. 


shannonbaker:
Double tap the menu button to cycle the goggle input to the next input source. This may be used in conjunction with a new menu setting to define the quick switch sources to reduce the cycling options if the pilot only uses 2 inputs regularly.  Otherwise cycle through HDZ/HDMI/Expansion Module/AV.

Optionally, cycle through only detected inputs, but unsure if the goggle is able to detect all active/connected inputs.

shannonbaker:
Perhaps have a menu option to alter the jog wheel function from switching channels to switching input sources.

uPz3TT3r:
+vote for this! 
Alternative: option in menu to change input with function button instead of changing fan speed.

bkleiner:
Lets keep the interactions well defined. Double tap cycles through all import sources.

a-di-ez:
@bkleiner I think it's better to double-click on the OK button of the wheel to call a pop-up window with a list of all sources and select the desired one with the wheel, then after a pause of 2 seconds automatically or by pressing OK - select this channel.

This is how we get rid of:
- the need to scan only connected sources
- from accidentally switching the source instead of entering the menu
- accidentally pressing twice, which will cause the source to switch, just wait a couple of seconds or press the OK button on the wheel and the selection menu disappears

And: 
- it's faster to select only what you need than to jump over all sources

Similarly, it would be convenient to adjust the fan speed: we hold the button for a couple of seconds and the menu for changing the fan speed from 1 to 10 appears, since now the most minimal fan mode creates a lot of noise and a couple of points below are missing to make the glasses silent. Plus it will speed up the fan speed reset after we blew the lenses, instead of switching all the speeds in a circle.

What do you think?

bkleiner:
> @bkleiner I think it's better to double-click on the OK button of the wheel to call a pop-up window with a list of all sources and select the desired one with the wheel, then after a pause of 2 seconds automatically or by pressing OK - select this channel.
> 
> This is how we get rid of:
> 
> * the need to scan only connected sources
> * from accidentally switching the source instead of entering the menu
> * accidentally pressing twice, which will cause the source to switch, just wait a couple of seconds or press the OK button on the wheel and the selection menu disappears
> 
> And:
> 
> * it's faster to select only what you need than to jump over all sources
> 
[...]

PR #94, opened 2023-01-17
Implement Head Tracking

pkendall64:
Implementation of Head Tracking.

The head tracker in the HDZero goggles outputs the PPM signal on the HT port on the goggles with the following channel layout
- Pan -> Channel 0 (TR1 in EdgeTX)
- Tilt -> Channel 1 (TR2 in EdgeTX)
- Roll - Channel 2 (TR3 in EdgeTX)

## Setup
In the goggles "Head Tracking" page the "Tracking" option should be set to "On", the the goggles should be resting on a level surface and the "Calibrate" option should be selected. I find that resting them on a book on the left and right side, with the face plate above the desk gives the best results.

The "Max Angle" option is used to set the angular limits for the head tracker, i.e. how far your head moves to reach full deflection on the PPM signal. The best way to do this is to first get you head in natural resting position when flying and set the centre point using the "Set Center" option, then select the "Max Angle" option and adjust the value while moving your head left/right, up&down until you get the desired range on the PPM bars at the bottom of the screen.

## Usage
When you use the head tracking function, you need to enter the "Head Tracking" page and select the "Set Center" option. This will reset the centre/home point for the goggles, so you need to do this when facing the direction and holding you head in you natural resting position when flying.

### DIFF
--- a/src/core/ht.c
+++ b/src/core/ht.c
@@ -14,13 +14,12 @@
 #include "ht.h"
 #include "osd.h"
 #include "MadgwickAHRS.h"
-#include "../bmi270/accel_gyro.h"
-#include "../driver/hardware.h"
-#include "../driver/dm6302.h"
-#include "../driver/oled.h"
-#include "ui/page_common.h"
-#include "ui/page_scannow.h"
 
+#include "bmi270/accel_gyro.h"
+#include "driver/hardware.h"
+#include "driver/dm6302.h"
+#include "driver/oled.h"
+#include "ui/page_common.h"
 
 //#define FAST_SIM
 ///////////////////////////////////////////////////////////////////////////////
@@ -28,9 +27,16 @@
 static ht_data_t ht_data;
 static uint8_t frame_period = 10;
 static uint8_t sync_len = 200;
+
 static volatile bool calibrating = false;
 static int calibration_count = 0;
-static float imu_rotation[3] = {0.0 * DEG_TO_RAD, -90.0 * DEG_TO_RAD, (-90.0+22.0) * DEG_TO_RAD};
+
+static const float imu_orientation[3] = {0.0 * DEG_TO_RAD, -90.0 * DEG_TO_RAD, (-90.0+22.0) * DEG_TO_RAD};
+
+static const int ppmMaxPulse = 500;
+static const int ppmMinPulse = -500; 
+static const int ppmCenter = 1500; 
+
 
 ///////////////////////////////////////////////////////////////////////////////
 //no motion to disable OLED display
@@ -142,7 +148,7 @@ void get_imu_data(int bCalcDiff)
     }
 }
 
-static void thread_imu(union sigval timer_data)
+static void timer_callback_imu(union sigval timer_data)
 {
     get_imu_data(true);
     calc_ht();
@@ -159,16 +165,6 @@ void init_ht()
     ht_data.tiltInverse = 1; 
     ht_data.rollInverse = -1; 
     ht_data.panInverse = -1; 
-
-    ht_data.tiltMaxPulse = 500;
-    ht_data.tiltMinPulse = -500; 
-    ht_data.tiltCenter = 1500; 
-    ht_data.panMaxPulse = 500;
-    ht_data.panMinPulse = -500; 
-    ht_data.panCenter = 1500; 
-    ht_data.rollMaxPulse = 500;
-    ht_data.rollMinPulse = -500; 
-    ht_data.rollCenter = 1500; 
 
     ht_data.htChannels[0] = 0;
     ht_data.htChannels[1] = 0;
@@ -193,7 +189,7 @@ void init_ht()
                             };
 
     sev.sigev_notify = SIGEV_THREAD;
-    sev.sigev_notify_function = &thread_imu;
+    sev.sigev_notify_function = &timer_callback_imu;
 
     int res = timer_create(CLOCK_REALTIME, &sev, &timerId);
     if (res != 0){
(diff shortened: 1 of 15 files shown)


src/core/ht.c
```diff
@@ -135,136 +148,147 @@ void get_imu_data(int bCalcDiff)
...
+    sev.sigev_notify_function = &timer_callback_imu;
+
+    int res = timer_create(CLOCK_REALTIME, &sev, &timerId);
+    if (res != 0){
+        LOGE("Error timer_create: %s\n", strerror(errno));
+        return;
+    }
 
-    accAngle[2] = acc_to_g(ht_data.sensor_data.acc.z - ht_data.acc_offset[2]);
+    res = timer_settime(timerId, 0, &its, NULL);
```
bkleiner, line 200: lovely. we should consider adopting this same pattern for the other interval based threads.

src/core/ht.c
```diff
@@ -135,136 +148,147 @@ void get_imu_data(int bCalcDiff)
...
+static void calc_acc(float* accAngle) //in G
+{
+    // convert accelerometer readings to G forces
+    accAngle[0] = acc_to_g(ht_data.sensor_data.acc.x);
+    accAngle[1] = acc_to_g(ht_data.sensor_data.acc.y);
+    accAngle[2] = acc_to_g(ht_data.sensor_data.acc.z);
+    rotate(accAngle, imu_orientation);
+}
 
+static int constrain(int value, int min, int max)
```
bkleiner, line 261: shall we move these to a dedicated subdir eg `util`?

src/core/ht.c
```diff
@@ -135,136 +148,147 @@ void get_imu_data(int bCalcDiff)
...
+    if (value < min)
+        return min;
+    if (value > max)
+        return max;
+    return value;
+}
+
+// Normalizes any number to an arbitrary range
+// by assuming the range wraps around when going below min or above max
+static float normalize(float value, float start, float end)
```
bkleiner, line 272: dito.

src/core/ht.c
```diff
@@ -273,55 +297,69 @@ void calibrate_ht()
...
+	MadgwickAHRSupdateIMU(gyrAngle[0] * DEG_TO_RAD, gyrAngle[1] * DEG_TO_RAD, gyrAngle[2] * DEG_TO_RAD, \
+                          accAngle[0],              accAngle[1],              accAngle[2]);
 
-    iir_filter(ht_data.tiltRollBeta,ht_data.tiltAngle, &ht_data.tiltAngleLP);
-    iir_filter(ht_data.tiltRollBeta,ht_data.rollAngle, &ht_data.rollAngleLP);
-    iir_filter(ht_data.panBeta, normalize(ht_data.panAngle, -180,180), &ht_data.panAngleLP);
+    // Adjust PTR relatice to user specified home position
+	ht_data.panAngle = getYaw() - ht_data.panAngleHome;
+	ht_data.tiltAngle = getPitch() - ht_data.tiltAngleHome;
+	ht_data.rollAngle = getRoll() - ht_data.rollAngleHome;
```
bkleiner, line 341: whitespace

src/core/ht.c
```diff
@@ -273,55 +297,69 @@ void calibrate_ht()
...
-
-    tmp = ht_data.rollAngleLP * ht_data.rollInverse * ht_data.rollFactor;
-    if((tmp > ht_data.rollMinPulse) && (tmp < ht_data.rollMaxPulse)) {
-        tmp += ht_data.rollCenter;
-        ht_data.htChannels[2] = (int16_t)tmp;
-    }
-*/
-    //Set_HT_dat(1500, 1500, 1500);
+    Set_HT_dat(ht_data.htChannels[0], ht_data.htChannels[1], ht_data.htChannels[2]);
 	return 1;
```
bkleiner, line 353: whitespace

bkleiner:
erff, seems github is refusing to rebase because of "too many changes".
could you rebase manually when you get the chance?

src/core/ht.c
pkendall64: So you want me to create a `util` folder and have a `math.c` & `math.h` file?
We could mode `normalize`, `constrain` to inline functions in the .h and move `rotate` to the .c file.

pkendall64:
> erff, seems github is refusing to rebase because of "too many changes". could you rebase manually when you get the chance?

Rebased and pushed, with whitespace fixes.

src/core/ht.c
bkleiner: yeah and probably `file_compare` and `file_exists` into something like `file.{c,h}`
for constrain a macro maybe more suitable tho, cover all the types.
i can submit a PR on top of this one, once its merged.

pkendall64: Yeah, maybe a seperate PR for extracting utility functions and macros might be better

pkendall64: I caved and did the refactoring to extract those functions 😃

bkleiner: happens to the best of us :)

Yith1:
I've not had chance to try this functionality but I noticed something in the description I thought worth commenting on.

> When you use the head tracking function, you need to enter the "Head Tracking" page and select the "Set Center" option. This will reset the centre/home point for the goggles, so you need to do this when facing the direction and holding you head in your natural resting position when flying.

This is good, but ideally you'd want a quick way to do this reset as you might have to do it mid flight. Maybe a short click on the right button of the goggles if the Headtracker is enabled?

bkleiner:
_As far as i can tell_, this works nicely. The bars on the tx follow the goggles movement. Would be good to have somebody with an gimbal try it too tho.

For button interaction: long press of the right button is currently used to modify the top fan speed (quick de-fog).
i think in the long run we gonna have to implement more gestures like double tap to cover all the functionality. 

Feel like it might be a better idea to get this merged and take stock of the button interactions in another PR potentially also implementing #27

Yith1:
My test of the Headtracker. In conclusion it worked fine. 
Only minor gripe would be that it was easy to click "calibrate" instead of "centre" when picking up the goggles and senterring them ready for use. Perhaps once calibrated that could be the first option? However I'm not sure if moving things around like that is a good user-experience.

PR #107, opened 2023-01-20
Head tracking cleanup

pkendall64:
Just some cleanup so we can avoid issues later on

### DIFF
--- a/src/core/ht.h
+++ b/src/core/ht.h
@@ -7,8 +7,8 @@
 #define CALIBRATION_BCNT   8 // calibartion loop cnt = (1<<CALIBRATION_BCNT)
 #define gyroWeightTiltRoll 0.98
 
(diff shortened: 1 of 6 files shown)


src/core/ht.h
```diff
@@ -34,28 +34,24 @@ typedef struct {
...
-	
-} ht_data_t;
 
+} ht_data_t;
 
 void init_ht();
 void enable_ht();
 void disable_ht();
 void calibrate_ht();
 void set_maxangle_ht(int angle);
```
bkleiner, line 53: _personally_ i prefer naming functions `<subject>_<action>` eg, `ht_init` vs `init_ht`
this seems also be the case in a good portion of the codebase. 
can we adjust these function names?

pkendall64: Agreed, and done. This also makes it more inline with all the other code.

PR #109, opened 2023-01-22
Antenna OSD order

shannonbaker:
Antenna glyph ordering confusing and changed to left to right 1, 2, 3, 4 when viewing goggles top down.


Schlaegel:
I'm not sure what you mean: "When viewing goggles top down"

Is it the order of the antenna from the perspective of the wearer when wearing the goggles? I hope so.

I recommend bumping up the graphic for the higher sma so it's clear which are which.

shannonbaker:
> I'm not sure what you mean: "When viewing goggles top down"
> 
> Is it the order of the antenna from the perspective of the wearer when wearing the goggles? I hope so.
> 
> I recommend bumping up the graphic for the higher sma so it's clear which are which.

image

Issue #64, opened 2023-01-10
Devcontainer not working on M1 Mac


creecedev:
When starting the remote dev container, the setup.sh runs and after the media is extracted, there's a CMake error that a simple test program cannot be compiled. The same process of cloning into the container and building on my Windows machine had no issues.

I am new to Docker in general but i'm doing the same thing on both systems and getting different results.

C++/CMake/CMake Tools are installed as extensions within the dev container.

[variant] Loaded new set of variants
[proc] Executing command: /usr/bin/aarch64-linux-gnu-gcc -v
[proc] Executing command: /usr/bin/aarch64-linux-gnu-gcc-11 -v
[proc] Executing command: /usr/bin/c89-gcc -v
[proc] Executing command: /usr/bin/c99-gcc -v
[proc] Executing command: /usr/bin/gcc -v
[proc] Executing command: /usr/bin/gcc-11 -v
[proc] Executing command: /bin/aarch64-linux-gnu-gcc -v
[proc] Executing command: /bin/aarch64-linux-gnu-gcc-11 -v
[proc] Executing command: /bin/c89-gcc -v
[proc] Executing command: /bin/c99-gcc -v
[proc] Executing command: /bin/gcc -v
[proc] Executing command: /bin/gcc-11 -v
[proc] Executing command: /bin/c99-gcc -v
[proc] Executing command: /bin/c89-gcc -v
[proc] Executing command: /usr/bin/c99-gcc -v
[proc] Executing command: /usr/bin/c89-gcc -v
[kit] Found Kit: GCC 11.3.0 aarch64-linux-gnu
[kit] Found Kit: GCC 11.3.0 aarch64-linux-gnu
[kit] Found Kit: GCC 11.3.0 aarch64-linux-gnu
[kit] Found Kit: GCC 11.3.0 aarch64-linux-gnu
[kit] Found Kit: GCC 11.3.0 aarch64-linux-gnu
[kit] Found Kit: GCC 11.3.0 aarch64-linux-gnu
[kit] Found Kit: GCC 11.3.0 aarch64-linux-gnu
[kit] Found Kit: GCC 11.3.0 aarch64-linux-gnu
[kit] Found Kit: GCC 11.3.0 aarch64-linux-gnu
[kit] Found Kit: GCC 11.3.0 aarch64-linux-gnu
[kit] Found Kit: GCC 11.3.0 aarch64-linux-gnu
[...]

PR #111, opened 2023-01-22
simplify mkapp

bkleiner:
Currently the build process generates allwinner's "variant" of an MBR, but does not package this MBR and only uses it to extract the app size from the specification ini. The binaries seem to be [crash prone](https://github.com/hd-zero/hdzero-goggle/issues/64).
This PR removes the sunxi-mbr tools and hardcodes the app size into the mkapp script directly.


mkapp/mkapp_ota.sh
```diff
@@ -1,128 +1,39 @@
...
-}
+make_img_md5 ${APP_IMAGE}
 
-envsetup
-make_app_res $SYS_PARTITION
-make_ota_package_for_app
\ No newline at end of file
+pushd $IMG_DIR
+tar cvf $IMG_DIR/hdzgoggle_app_ota-${APP_VERSION}.tar *
+popd
```
pkendall64, line 39: Probably better as
```
rm -f $IMG_DIR/hdzgoggle_app_ota-${APP_VERSION}.tar
tar cf $IMG_DIR/hdzgoggle_app_ota-${APP_VERSION}.tar -C $IMG_DIR .
```

mkapp/mkapp_ota.sh
```diff
@@ -1,128 +1,39 @@
 #!/bin/bash
+set -x
```
pkendall64, line 2: Probably remove this as it make the build a bit noisy

bkleiner: ....left over from debugging. tho a `set -e` might make sense actually.

mkapp/mkapp_ota.sh
bkleiner: i had something like that initially, but that leaves us with an ugly leading `./` in the archive, which is a bit of a pet-peeve of mine. Most tools should handle that fine, but hard to tell if that would cause any issues down the line.

pkendall64: Fair call. I just don't think the `popd` is required because the script is ending. Which means the `pushd` can just be `cd`.

mkapp/mkapp_ota.sh
pkendall64: Yeah `set -e` would be better.

PR #113, opened 2023-01-23
Embedded OSD mode to include 16x9 width.

shannonbaker:


src/core/osd.c
```diff
@@ -335,7 +335,25 @@ void osd_channel_show(bool bShow)
 static void create_osd_object(lv_obj_t **obj, const char *img, int index)
 {
 	int x = 0;
+	int left_offset = 0;
+	int right_offset = 0;
+	int osd_embedded_mode;
+	char str[128];
+	
+	ini_gets("osd", "embedded_mode", "2", str, sizeof(str), SETTING_INI);
+	osd_embedded_mode = atoi(str);
```
pkendall64, line 344: It would be better adding `embedded_mode` to `g_setting` structure and loading this one time in `main.c` where all the other initialisation settings are loaded. Also use `ini_getl` to load a number setting.

shannonbaker: moved but leaving as string for now as I may change values to "4x3" and "16x9" respectively.

src/core/main.c
```diff
@@ -74,6 +74,10 @@ static void load_ini_setting(void) {
     }
     g_setting.autoscan.source = ini_getl("autoscan", "source", SETTING_SOURCE_LAST, SETTING_INI);
     g_setting.autoscan.last_source = ini_getl("autoscan", "last_source", SETTING_SOURCE_HDZERO, SETTING_INI);
+	
+	//embedded osd mode 4x3 default
+	ini_gets("osd", "embedded_mode", "2", str sizeof(str), SETTING_INI);
+	g_setting.osd.embedded_mode = atoi(str);
```
pkendall64, line 81: Indentation should be spaces. The project has a `.clang-format` file so any new changes should follow that format.

Also, you have not defined an `osd` structure with `embedded_mode` as a field in `page_common.h`, so this code will not compile as it is.

Perhaps `embedded_mode` is not the right name, maybe `aspect_ratio` would be better?

src/ui/page_imagesettings.c
```diff
@@ -52,8 +53,9 @@ static lv_obj_t *page_imagesettings_create(lv_obj_t *parent, panel_arr_t *arr) {
     create_slider_item(&slider_group2, cont, "Saturation", 47, g_setting.image.saturation, 2);
     create_slider_item(&slider_group3, cont, "Contrast", 47, g_setting.image.contrast, 3);
     create_slider_item(&slider_group4, cont, "OLED Auto off", 3, g_setting.image.auto_off, 4);
+    create_slider_item(&slider_group5, cont, "Embedded OSD Mode", 2, g_setting.osd.embedded_mode, 5);
```
pkendall64, line 56: I think the `2` should be `1` as the function uses that as the `max` value with `0` being the minimum.
So on the main display when it's 16x9 the slider is in the middle!

pkendall64 (commented):
I wonder whether `embedded_mode` would be better named `aspect_ratio`.

ligenxxxx:
- Can the Auto option be added?
- I don't think it's appropriate to put it in the image setting submenu. Considering that there may be more OSD settings in the future, it would be better to add the osd submenu, although there may only be one setting for now.

bkleiner:
> I wonder whether `embedded_mode` would be better named `aspect_ratio`.

_personally_ i don't think so, as its not really a `aspect_ratio` either, this setting controls a spacing, a placement, where the values happen to match the two different aspect ratios the goggles currently offer.

bkleiner:
> * Can the Auto option be added?

Yes, ultimately i would like to see a Auto option as well, however this PR was already getting way larger than expected so i omitted it for the time being. 

> * I don't think it's appropriate to put it in the image setting submenu. Considering that there may be more OSD settings in the future, it would be better to create a osd submenu, although there may only be one setting for now.

I agree, but it seemed silly to add a new menu for a single setting and the image-settings menu gave us i nice preview.
As soon as we have a second setting to adjust, i too am in favor off adding a new menu entry.

PR #122, opened 2023-01-26
Implement manual cell count mode

Nikolas-S:
Added a toggle to the power menu to switch between manual and automatic cell count mode, as well as a slider to set the cell count. 
In manual mode, the value from the slider is used. 
In automatic, the existing method is used to determine the cell count automatically and the slider value is ignored.
Also moved the value init of the cell voltage slider and warning type button group from load_ini_setting() in main.c to the end of page_power_create() in page_power.c and replaced a few hardcoded values with defines.


mkapp/app/setting.ini
```diff
@@ -15,6 +15,8 @@ source=0
 [power]
 voltage=35
 warning_type=2
+s_count_mode=0
+s_count = 2
```
bkleiner, line 19: let's please go with some more descriptive names here eg. `cell_cout`

src/core/defines.h
```diff
@@ -9,7 +9,8 @@ typedef enum{
     OPLEVEL_IMS = 11,
     PAGE_FAN_SLIDE = 100,
     PAGE_ANGLE_SLIDE = 101,
-    PAGE_POWER_SLIDE = 102,
+    PAGE_POWER_SLIDE_S_COUNT = 102,
```
bkleiner, line 12: dito

src/ui/page_common.h
```diff
@@ -89,6 +89,8 @@ typedef struct {
     int voltage;
     bool display_voltage;
     int warning_type; // 0=beep,1=visual,2=both
+    int s_count_mode; // 0=auto,1=manual
+    int s_count;
```
bkleiner, line 93: dito

src/ui/page_power.c
```diff
@@ -42,48 +44,103 @@ static lv_obj_t *page_power_create(lv_obj_t *parent, panel_arr_t *arr) {
...
+    btn_group_set_sel(&btn_group_s_count_mode, g_setting.power.s_count_mode);
+    lv_slider_set_value(slider_group_s_count.slider, g_setting.power.s_count, LV_ANIM_OFF);
+    lv_slider_set_value(slider_group_c_voltage.slider, g_setting.power.voltage, LV_ANIM_OFF);
+    btn_group_set_sel(&btn_group_warn_type, g_setting.power.warning_type);
+    set_battery_S();
 
     return page;
 }
 
 void set_battery_S() {
```
bkleiner, line 76: dito

src/ui/page_power.c
```diff
@@ -42,48 +44,103 @@ static lv_obj_t *page_power_create(lv_obj_t *parent, panel_arr_t *arr) {
...
+    }
+
     sprintf(str, "%dS", g_battery.type);
-    lv_label_set_text(label_s, str);
+    lv_label_set_text(label_cell_count, str);
 }
 
-void set_voltage(int val) {
-    lv_slider_set_value(slider_group.slider, val, LV_ANIM_OFF);
+void power_s_count_inc(void) {
```
bkleiner, line 94: dito

src/ui/page_power.h
```diff
@@ -1,6 +1,18 @@
...
 #define _PAGE_POWER_H
 
+#define S_MIN_COUNT 2
+#define S_MAX_COUNT 5
+#define CELL_VOLTAGE_MIN 28
+#define CELL_VOLTAGE_MAX 42
+
+#define ROW_BATT_C_LABEL 0
+#define ROW_S_COUNT_MODE 1
+#define ROW_S_COUNT 2
```
bkleiner, line 11: dito

bkleiner (changes requested):

bkleiner:
just gave it a quick test, appears we need to remove main.c line 192.
that function is called after the ui init and thus the auto-detected value gets overwritten. 
lets also up the max cell count to 6, with li-fe and li-ion batteries 6s is a safe configuration.

Nikolas-S:
Was originally going with s_count because if you have two cells in parallel the cell count would be 2 but the battery would be 1S, cell_count is definitely more intuitive though 😄 
Also upped max cell count to 6 and removed line 192 from main.c

bkleiner (changes requested):
neat! lets address the enum, but otherwise looks good now.

bkleiner:
oh and `page_power.c` line 210 needs to be adjust to 6; we have 2 new elements on that page now.

Nikolas-S:
Good call on the max element count, didn't even realize I had to adjust that

bkleiner:
awesome! LGTM now, maybe @evilC can give it a try as well.

bkleiner (approved):

evilC:
I'll try and get around to it tomorrow

evilC:
Seems to work, but I have two observations:

1. For the cell count slider, rotating the dial clockwise decreases the cell count, and anti-clockwise increases. Should this not be the other way around?
2. I am not sure if this is related to your changes, but the warning does not kick in when the voltage reaches the specified level, it kicks in when it falls below it.
ie I have warning set to 2.8v per cell, and cell count set to 4. I would expect the warning to kick in when the voltage reaches 11.2v, but it does not kick in until it reaches 11.1v

Nikolas-S:
> Seems to work, but I have two observations:
> 
>     1. For the cell count slider, rotating the dial clockwise decreases the cell count, and anti-clockwise increases. Should this not be the other way around?
> 
>     2. I am not sure if this is related to your changes, but the warning does not kick in when the voltage reaches the specified level, it kicks in when it falls below it.
>        ie I have warning set to 2.8v per cell, and cell count set to 4. I would expect the warning to kick in when the voltage reaches 11.2v, but it does not kick in until it reaches 11.1v

Regarding 1, I implemented it so that the increase direction is the same as for the cell voltage, if the directions are different between the cell voltage and cell count slider there's a bug somewhere, if not, maybe it's worth thinking about a feature that lets the user set the direction they have to rotate in in order to increase/decrease values.

Regarding 2, that's not something I changed, as far as I can tell there's two places in the code where the cell voltage is checked and in both places the voltage needs to be lower than the set voltage.
Once in ui_statusbar.c: https://github.com/hd-zero/hdzero-goggle/blob/main/src/ui/ui_statusbar.c#L162
And once inosd.c: https://github.com/hd-zero/hdzero-goggle/blob/main/src/core/osd.c#L217

evilC:
> Regarding 1, I implemented it so that the increase direction is the same as for the cell voltage, if the directions are different between the cell voltage and cell count slider there's a bug somewhere, if not, maybe it's worth thinking about a feature that lets the user set the direction they have to rotate in in order to increase/decrease values.

You are indeed correct. I had not noticed it before, and now I have, I cannot un-notice it. I am guessing this is a standard slider control that is just re-used throughout the UI and it's how that control has been coded. Raised #124 as I think this should be changed

> Regarding 2, that's not something I changed, as far as I can tell there's two places in the code where the cell voltage is checked and in both places the voltage needs to be lower than the set voltage. Once in ui_statusbar.c: https://github.com/hd-zero/hdzero-goggle/blob/main/src/ui/ui_statusbar.c#L162 And once in osd.c: https://github.com/hd-zero/hdzero-goggle/blob/main/src/core/osd.c#L217

Raised #125

evilC:
Oh, and one more thing.  
I see "Auto" mode has been preserved - and I am guessing that your code has not changed the fact that the warning code is fundamentally broken in Auto mode? (ie it mis-interprets a 4S 18650 pack at 3v per cell as a 3S pack, and does not warn you of low voltage)  
So I am guessing I should raise another issue to request that when in auto mode, the `Cell Voltage` option should be replaced by a `Pack Voltage` option.  

ie
In Manual mode, you set the cell count and the minimum voltage per cell
In Auto mode, you set the minimum voltage of the supply - it does not try to calculate the cell voltage

Nikolas-S:
> Oh, and one more thing.  
> I see "Auto" mode has been preserved - and I am guessing that your code has not changed the fact that the warning code is fundamentally broken in Auto mode? (ie it mis-interprets a 4S 18650 pack at 3v per cell as a 3S pack, and does not warn you of low voltage)  
> So I am guessing I should raise another issue to request that when in auto mode, the `Cell Voltage` option should be replaced by a `Supply Voltage` option.  
> 
> ie
> In Manual mode, you set the cell count and the minimum voltage per cell (eg 2.8v)
> In Auto mode, you set the minimum voltage of the supply (eg 12v) - it does not try to calculate the cell voltage
> 
> 

Nope, didn't change anything about the way auto mode works, I just added the option to set the cell count manually.
Might be worth a discussion whether to keep calling it auto/manual mode at that point, or have something like "per cell" and "total".

bkleiner:
@evilC can we please stick to the matters a given PR actually modifies.
i gather there are not further notes on the things that actually did change here.

PR #132, opened 2023-01-28
move hot clips to end of list

bkleiner:


evilC:
Flashed [this version](https://github.com/hd-zero/hdzero-goggle/actions/runs/4033844486) to my goggles. Will be testing it properly when I fly tomorrow, but looking good on the bench

pkendall64:
I was just about so say you could allow the user to choose the sort order in the goggles, but you merged it 😆

bkleiner:
> I was just about so say you could allow the user to choose the sort order in the goggles, but you merged it laughing

yeah, there was a side-discussion on discord, where we came to the conclusion that having the hot clips in an seperate tab or section would be ultimately preferable. this is the "botched" workaround until that is realized.

PR #134, opened 2023-01-29
Adopt new versioning scheme

bkleiner:
This PR proposes a new versioning scheme, where the goggle application gets a full semver,
however the HAL versions are preserved:
`<RX_VER>-<VA_VER>-<APP_SEMVER>`

so the current version would become:
`HDZERO_GOGGLE-68-127-7.1.0.bin`

additionally development versions (eg. versions not built off a git tag) append the their current commit hash:
`HDZERO_GOGGLE-68-127-7.1.0-d1f1b02.bin`


pkendall64:
Works for me. What do you think about moving the version to be on the same line as the "Update Goggle" similar to the "Update ESP32"?
May be it could be split out too, `ver: 7.1.0, rx: 68, va: 127`

Yith1:
> > Works for me. What do you think about moving the version to be on the same line as the "Update Goggle" similar to the "Update ESP32"? May be it could be split out too, `ver: 7.1.0, rx: 68, va: 127`
> 
> i like that, i will look into it.

I'm still unsure what the different version numbers actually mean.
rx: is that the divimath number?
ver: is that the linux based firmware?
va: I have no clue about at all...

bkleiner:
rx and va are the divimath receiver and fpga firmware versions respectively.
they are important for divimath to troubleshoot reception and image processing issues.

ver would describe the version of the goggles main application, developed in this repository.

underlying linux system is another issue yet, as there is only one version out in the wild,
we should be fine on that for the time being, but i would like to version that eventually too.

bkleiner:
> Works for me. What do you think about moving the version to be on the same line as the "Update Goggle" similar to the "Update ESP32"? May be it could be split out too, `ver: 7.1.0, rx: 68, va: 127`

humm, turns out divimath sneakily hid a method for writing to the DM5680 registers behind the "Current Version" label.
Might or might not be important, but i don't want to just delete it.

PR #143, opened 2023-01-30
Fix self test logging

bkleiner:
Carl made me aware that the "self test" feature stopped working. That previously hooked into their flavor of `Printf` and was lost during the log refactor.
This PR re-adds this functionality.


ligenxxxx:
I test it with no problem.

PR #149, opened 2023-02-01
Fixes OLED wake from motion detect

pkendall64:
There was an intermittent problem when the OLED off/on is called from the timer thread where the OLED on would not always work causing the screens to stay off.

This PR changes that behaviour by just setting a flag in the timer thread, and the `detect_motion` call is now done from the main thread which perform the OLED changes if there was motion or not.

Prior to this fix I could easily get the OLED wake to fail within a couple of tries, but after this I have performed the sleep/wake well over 10 times without a problem.

### DIFF
--- a/src/core/ht.c
+++ b/src/core/ht.c
@@ -28,7 +28,7 @@ static const uint8_t frame_period = 10;
 static const uint8_t sync_len = 200;
 
 static bool has_motion_data = false;
-static int is_moving;
+static bool is_moving = true;
 
 
 static volatile bool calibrating = false;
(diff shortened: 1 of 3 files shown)


src/core/ht.c
```diff
@@ -27,6 +27,10 @@ static ht_data_t ht_data;
 static const uint8_t frame_period = 10;
 static const uint8_t sync_len = 200;
 
+static bool has_motion_data = false;
+static int is_moving;
```
bkleiner, line 31: can we make this a `bool`? its exclusively used as one.
dito in as `detect_motion` parameter

bkleiner:
lovely, i could not provoke a fault once in about 10 attempts. prior i would hit it ~3 attempts in.

pkendall64:
I've cherry-picked the cleanup code from your repository @bkleiner and retested again 10 times without issue.

PR #150, opened 2023-02-02
add record app sources

bkleiner:
Carl sent them over earlier today. PR crudely integrates them into the the tree.
I ran a quick test - application seems to be operating as expected.


bkleiner:
merging this so its available for everyone to play with

PR #153, opened 2023-02-05
add semaphore to trigger osd re-render

bkleiner:
This PR introduces a semaphore to signal the osd thread to re-render.
this significantly reduces cpu load. (about 40% according to `top`)
first two commits do not contain any functional changes, they are purely cosmetic.


src/core/msp_displayport.c
```diff
@@ -440,11 +439,11 @@ void clear_screen() {
...
 
 void update_osd(uint16_t *line_buf, uint8_t row) {
-    uint8_t i;
-
-    for (i = 0; i < HD_HMAX; i++) {
+    for (uint8_t i = 0; i < HD_HMAX; i++) {
         if (osd_buf[row][i] != line_buf[i]) {
             osd_buf[row][i] = line_buf[i];
         }
     }
```
pkendall64, line 446: Why not replace the entire for loop with the following? 
```
memcpy(osd_buf[row], line_buf, HD_HMAX);
```

pkendall64 (commented):
Looks good, but I'm unable to test for the next few days.

alfskaar (approved):

PR #162, opened 2023-02-09
Add option to display average cell voltage in osd

### LINKED ISSUE #68: Feature request: Option to show average cell voltage
Please add an equivalent of BetaFlight's average cell voltage  
If you implemented #67 , then this would always be accurate-ish

ie, if (Once #67 is implemented) I have set the cell count set to 4S, and the current voltage of the pack is 12v, I want to see `3v` in the OSD for the voltage

Nikolas-S:
Added a toggle to the power settings page to choose whether the osd shows the total battery voltage or the average voltage per cell.
I wanted to make it obvious which mode is active, so "/C" is added to the old format when cell average mode is active, if someone has a better idea let me know.
e.g.:
"Total" mode:       4S 15.6V
"Cell Avg." mode: 4S 3.9V/C

Would appreciate if someone could test the pr on their goggles, don't have mine yet.

Resolves #68

### DIFF
--- a/src/ui/ui_statusbar.c
+++ b/src/ui/ui_statusbar.c
@@ -151,27 +151,28 @@ void statubar_update(void) {
     memset(buf, 0, sizeof(buf));
 
     // display battery voltage
-    int bat_mv = 0;
-    switch (g_setting.power.osd_display_mode)
-    {
-        default:
-        case SETTING_POWER_OSD_DISPLAY_MODE_TOTAL:
-            bat_mv = battery_get_millivolts(false);
-            sprintf(buf, "%dS %d.%dV", 
-                    g_battery.type,
-                    bat_mv / 1000,
-                    bat_mv % 1000 / 100);
-            
-            break;
-            
-        case SETTING_POWER_OSD_DISPLAY_MODE_CELL:
-            bat_mv = battery_get_millivolts(true);
-            sprintf(buf, "%dS %d.%dV/C", 
-                    g_battery.type,
-                    bat_mv / 1000, 
-                    bat_mv % 1000 / 100);
+    switch (g_setting.power.osd_display_mode) {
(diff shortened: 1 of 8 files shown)


src/ui/ui_statusbar.c
```diff
@@ -149,9 +149,30 @@ int statusbar_init(void) {
...
     char buf[128];
     memset(buf, 0, sizeof(buf));
-    sprintf(buf, "%dS %d.%dV", g_battery.type,
-            g_battery.voltage / 1000,
-            g_battery.voltage % 1000 / 100);
+
+    // display battery voltage
+    int bat_mv = 0;
+    switch (g_setting.power.osd_display_mode)
+    {
```
bkleiner, line 156: looks like code-style does not align with the clang-format file. if you are using vscode `command palette -> format document` should fix that.

src/ui/ui_statusbar.c
```diff
@@ -149,9 +149,30 @@ int statusbar_init(void) {
...
+        default:
+        case SETTING_POWER_OSD_DISPLAY_MODE_TOTAL:
+            bat_mv = battery_get_millivolts(false);
```
bkleiner, line 159: declaring a scope around your case will allow you to use temporaries within the switch:
```c
case SETTING_POWER_OSD_DISPLAY_MODE_TOTAL: {
    const int bat_mv = battery_get_millivolts(false);
    [...]
    break;
}
```

bkleiner (commented):
Lovely! merging probably gotta wait until 8.0.0 is fully released.
i will give it a try on my goggles asap.

bkleiner:
Works great overall, but the menu option "OSD Display Mode" is getting cut off.
Can simply we shorten it to "Display Mode"?

PR #171, opened 2023-02-21
Add ELRS OSD and buzzer support

pkendall64:
Adds support for receiving MSP commands to add the the OSD overlay. And also implements the buzzer command over MSP as well. 

The current VTx OSD takes precedence, so if the ELRS elements are put in places that are used then the existing ones are displayed.

This is in preparation for other backpack features that are in the works.

To test you should flash https://github.com/ExpressLRS/Backpack/pull/84 to the backpack. Also flash an ESP32 with the DEBUG_TX_Backpack and you can use the `osd_test.py` script in the Backpack repo for sending text to the goggle OSD.

### DIFF
--- a/src/core/elrs.c
+++ b/src/core/elrs.c
@@ -20,6 +20,7 @@
 #include "core/battery.h"
 #include "core/common.hh"
 #include "core/ht.h"
+#include "core/msp_displayport.h"
 #include "core/osd.h"
 #include "driver/dm5680.h"
 #include "driver/gpio.h"
@@ -42,7 +43,11 @@ static int record_state;
 static uint32_t record_time = 0;
 static bool headtracking_enabled = false;
 
+uint8_t elrs_osd[HD_VMAX][HD_HMAX];
+static uint8_t elrs_osd_overlay[HD_VMAX][HD_HMAX];
+
 void msp_process_packet();
+static void handleOSD(uint8_t *payload, uint8_t size);
 
 static const uint16_t freq_table[] = {
     5658, 5695, 5732, 5769, 5806, 5843, 5880, 5917, // R1-8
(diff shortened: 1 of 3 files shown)


src/core/elrs.c
```diff
@@ -42,7 +43,11 @@ static int record_state;
 static uint32_t record_time = 0;
 static bool headtracking_enabled = false;
 
+uint16_t elrs_osd[HD_VMAX][HD_HMAX];
+static uint16_t elrs_osd_overlay[HD_VMAX][HD_HMAX];
+
 void msp_process_packet();
+static void handleOSD(uint8_t *payload, uint8_t size);
```
bkleiner, line 50: code style, we are using `snake_case` elsewhere

src/core/elrs.c
```diff
@@ -364,4 +368,43 @@ void msp_ht_update(uint16_t pan, uint16_t tilt, uint16_t roll) {
 
 bool elrs_headtracking_enabled() {
     return headtracking_enabled;
+}
+
+void elrs_clear_osd() {
+    for (int i = 0; i < HD_VMAX; i++) {
```
bkleiner, line 374: maybe `memset` is more appropriate here.

bkleiner (changes requested):

src/core/elrs.c
pkendall64: Can't use memset because the underlying type is uint16_t

src/core/elrs.c
pkendall64: fixed

src/core/elrs.c
bkleiner: right, missed that

Issue #118, opened 2023-01-24
Feature Request: Show active Channel in HDZero OSD permanently


uPz3TT3r:
If you change a channel, the active channel is shown in OSD for some seconds, but goes away after that.
In Conditions with more pilots it is nice to know on which channel the goggle is set at the moment. It also helps in directly be able to see and fix a missconfiguration.

timfredriksson:
That sounds unnecessary as long as you can choose if you want it or not in the setting it seems fine

aviphysics:
What if it showed channel for a couple seconds whenever the goggle OSD was turned on, so you could always check your channel with at most a quick double click of the left button?

uPz3TT3r:
> 

@timfredriksson that would be perfect: a setting area in menü where you can turn off/on every OSD setting that you want permantly. Fan, Channel, Battery, Signal, Lock etc. etc. 

That would be nice.

PR #176, opened 2023-02-24
Add goggle battery voltage to inflight OSD

### LINKED ISSUE #169: Feature request : Display current goggle battery level (pack or cell) in the Flight/Video OSD
Please add the battery level to the in flight OSD so that the pilot is aware what the goggle battery level is while flying.

See https://github.com/hd-zero/hdzero-goggle/issues/68#issuecomment-1436723870

Nikolas-S:
Added the goggle battery voltage to the inflight OSD, as well as a toggle to show/hide it on the power settings page.
Voltage is shown in the same format as in the menu statusbar and turns red if the battery is low.

Gonna need some help testing this one since I don't have goggles yet, so there may be bugs.

Resolves #169


bkleiner:
i am kinda torn on this one, i think we want a dedicated menu to toggle osd elements and other options before we add anymore elements.
[there are likely a couple more elements](https://github.com/hd-zero/hdzero-goggle/issues/118) and we wanna be able to manage them, before we go nuts on these.

Zuldan:
@bkleiner totally agree. Being able to toggle all the OSD elements would be great and then eventually being able to position them.

Nikolas-S:
Yeah I was thinking the same thing, since there also already is also an option regarding the OSD in the image settings menu. (makes more sense than power menu but still)
But if the feature works, moving page entries to a different page wouldn't be much work, might do that next.

bkleiner:
> Yeah I was thinking the same thing, since there also already is also an option regarding the OSD in the image settings menu. (makes more sense than power menu but still) But if the feature works, moving page entries to a different page wouldn't be much work, might do that next.

yeah consens back then was that it wouldn't be worth adding a new menu for a single option, so once a dedicated osd menu is in place, that option should be moved.

Nikolas-S:
> > Yeah I was thinking the same thing, since there also already is also an option regarding the OSD in the image settings menu. (makes more sense than power menu but still) But if the feature works, moving page entries to a different page wouldn't be much work, might do that next.
> 
> yeah consens back then was that it wouldn't be worth adding a new menu for a single option, so once a dedicated osd menu is in place, that option should be moved.

Think we should close this pr for now or move the option later on once the OSD menu exists?

bkleiner:
> > > Yeah I was thinking the same thing, since there also already is also an option regarding the OSD in the image settings menu. (makes more sense than power menu but still) But if the feature works, moving page entries to a different page wouldn't be much work, might do that next.
> > 
> > 
> > yeah consens back then was that it wouldn't be worth adding a new menu for a single option, so once a dedicated osd menu is in place, that option should be moved.
> 
> Think we should close this pr for now or move the option later on once the OSD menu exists?

Let's convert it to a draft and rebase once the menu is in-place. changes look good overall.

Nikolas-S:
Sounds good 👍

Nikolas-S:
Might also be worth to refactor the inflight OSD a bit before we add more elements, especially the element positioning seems really jank. Wish I already had my goggles right about now, that's a bit beyond what I'll dare to put in a pr without any testing 😄

Schlaegel:
I think BetaFlight created a system for goggle side OSD elements that are enabled/placed by BetaFlight.

robhaswell:
My take is - don't let "great" be the enemy of "good".

My racing season has started and I'm already thinking that having the goggle battery shown would be a massive help. I don't want my warning to go off in the middle of a heat, but it's easy to forget to check during a busy race day.

I think we should test and merge this PR now, then clean up the OSD config at a later stage.

sercona:
feature that I'm asking for, that might be very related to this:

https://discord.com/channels/880782999006883870/880801151539347467/1097715740498472990

I'd like to be able to know how many flights I can get on a battery.  or how many minutes I can get.  this would mean that we'd have some way to zero a time counter, at will, and use that as our battery run time.  similarly, if we set a value and had it count down, it could give us a time estimate of when we are likely to run low (ie, an alarm or water mark for the alarm threshold).

SumolX:
Now that PR #222 has been merged in... should this PR be refactored to take into account #222?  Have it added and defaulted to hidden as part of the OSD UI Management?

Nikolas-S:
@SumolX Yup, I'll rebase. I think we should make it visible by default, I imagine a decent amount of people want the goggle voltage in their OSD and some might not realize it's an option it if it's hidden by default.

SumolX:
> @SumolX Yup, I'll rebase. I think we should make it visible by default, I imagine a decent amount of people want the goggle voltage in their OSD and some might not realize it's an option it if it's hidden by default.

I agree... I am one of them.

Nikolas-S:
Rebased after merge of #222

PR #182, opened 2023-02-27
refactor, add settings for osd element positioning

Nikolas-S:
Refactored OSD code to allow for individual element positioning and visibility settings (with settings saving x,y for both 4x3 and 16x9 modes) in preparation for a dedicated OSD settings page.
No (intentional) UI changes, besides changing the order of some OSD elements.
Added "g_setting_defaults" const setting_t struct instance to hold defaults for all settings values in one place.

Gonna need help testing this one, if the approach looks good.

### DIFF
--- a/mkapp/app/setting.ini
+++ b/mkapp/app/setting.ini
@@ -44,106 +44,106 @@ embedded_mode=0
 
 #### OSD elements ####
 # top fan speed
-element_show_topfan_speed=1
-element_pos_4_3_x_topfan_speed=160
-element_pos_4_3_y_topfan_speed=0
-element_pos_16_9_x_topfan_speed=0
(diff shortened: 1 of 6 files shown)


mkapp/app/setting.ini
```diff
@@ -41,3 +41,109 @@ enable=0
 
 [osd]
 embedded_mode=0
+
+#### OSD elements ####
+# top fan speed
+element_show_topfan_speed=1
```
bkleiner, line 47: nit-pick: could we go `element_<name>_show` etc? would make this a little more readable imho

bkleiner (commented):
nice. i like this approach.
PR tests fine, as far as i can tell, everything behaves just as it did before.

PR #184, opened 2023-02-28
refactor g_menu_op into dedicated file

bkleiner:


src/core/app_state.c
```diff
@@ -0,0 +1,137 @@
+#include "app_state.h"
+
+#include <log/log.h>
+#include <minIni.h>
+
```
pkendall64, line 5: need `#include <unistd.h>` for sleep, otherwise theres a warning.
Also, `input_device.h` needs `extern void exit_tune_channel();` to avoid a warning about that in this file too.

PR #186, opened 2023-02-28
refactor settings into dedicated file

bkleiner:


pkendall64 (commented):
Look good, compiles and runs without problems

PR #187, opened 2023-02-28
refactor dvr logic into dedicated file

bkleiner:


src/core/dvr.c
```diff
@@ -0,0 +1,142 @@
...
+        "in_linein"};
+
+    if (source > 2)
+        source = 2;
+    sprintf(buf, "%s %s", AUDIO_SEL_SH, audio_source[source]);
+    system(buf);
+}
+
+static void dvr_update_record_conf() {
+    if (g_setting.record.format_ts)
```
pkendall64, line 68: We should maintain a value with the values in it so we can short circuit this function if none of the values have changed. This *may* alleviate some of the problems seen with recordings flipping back to the defaults.

pkendall64 (commented):
This is a ver good idea to separate this code to it's own unit.

pkendall64 (commented):
This is a very good idea to separate this code to it's own unit.

As a side note:
`ui_main_menu.c` needs to include `dvr.h` other there are compile warnings

src/core/dvr.c
bkleiner: humm, while i do like the aspect of doing only as much as necessary, i have some concerns:
- if the dvr configuration reset is indeed related to writing the ini file, this will only make it more in-frequent and does not fix the root cause. we don't really want that right now, the issue is hard enough to trace as it is.
- if there are additional changes to the record parameters this could become very error prone because the developers will need to take great care that the file is indeed written out when one of the parameters change.

i vote we get this refactor merged and then investigate more. maybe take a peek at the minini write function, it does some interesting stuff with writing to a tmp file in-place and then moving it over. maybe we should swap that out for something where we have more control about when the actual file is written and, more importantly, we get notified if the write fails. right now there is zero validation whether a write actual took place or not.

pkendall64: Fair enough. If the minIni library was replaced with something where we have more control, then we could have that only do the writing if the data has changed.

PR #191, opened 2023-03-10
RTC Driver for setting/getting hardware clock.

SumolX:
Introducing RTC driver.  For setting up the initial date/time create a file on the root of the sd card called rtc.txt and add the following line to initialize the hardware clock:

2023-03-10T01:17:30

Once initialized the file can be deleted, otherwise the next time the goggles power up it will be 
set to the time specified in the file.  I purchased a compatible battery from Amazon:  https://a.co/d/ajZNTnZ

[rtc.txt](https://github.com/hd-zero/hdzero-goggle/files/10939234/rtc.txt)

### DIFF
--- a/src/driver/rtc.c
+++ b/src/driver/rtc.c
@@ -1,0 +1,129 @@
+#include "rtc.h"
+
+#include <linux/rtc.h>
+#include <sys/ioctl.h>
+#include <sys/time.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <unistd.h>
+
+#include <log/log.h>
+
+#include "util/file.h"
+
+/**
+ *  Constants
+ */
+static const char* RTC_FILE = "/mnt/extsd/rtc.txt";
+
+/**
+ *  Conversion Utils
+ */
+void rd2rt(const struct rtc_date* rd, struct rtc_time* rt) {
+	rt->tm_year = rd->year - 1900;
+	rt->tm_mon  = rd->month - 1;
+	rt->tm_mday = rd->day;
+	rt->tm_hour = rd->hour;
+	rt->tm_min  = rd->min;
+	rt->tm_sec  = rd->sec;
+}
+void rt2rd(const struct rtc_time* rt, struct rtc_date* rd) {
+	rd->year  = rt->tm_year + 1900;
+	rd->month = rt->tm_mon + 1;
+	rd->day   = rt->tm_mday;
+	rd->hour  = rt->tm_hour;
+	rd->min   = rt->tm_min;
+	rd->sec   = rt->tm_sec;
+}
+
+/**
+ *  Initialize RTC from file if detected,
+ *  must be formatted to ISO-8601:
+ *  YYYY-MM-DDTHH:MM:SS
+ *  2023-03-10T01:05:15
+ */
+void rtc_init() {
+	if (file_exists(RTC_FILE)) {
+		FILE* fp = fopen(RTC_FILE, "r");
+		if (fp) {
+			static char buffer[32];
+			while(fgets(buffer,sizeof(buffer),fp)) {
+				struct rtc_date rd;
+				sscanf(buffer,
+				       "%04d-%02d-%02dT%02d:%02d:%02d",
+					   &rd.year,
+					   &rd.month,
+					   &rd.day,
+					   &rd.hour,
+					   &rd.min,
+					   &rd.sec);
+			    rtc_set_clock(&rd);
+			}
+			fclose(fp);
+		}
+	}
+	
+	rtc_timestamp();
+}
+
+/**
+ *  Format RTC in a standard format
+ */
+static inline
+void rtc_print(const struct rtc_date* rd) {
+	LOGI("RTC: %04d-%02d-%02dT%02d:%02d:%02d\n",
+		 rd->year,
+		 rd->month,
+		 rd->day,
+		 rd->hour,
+		 rd->min,
+		 rd->sec);
+}
+
+/**
+ *  Log RTC in a standard format
+ */
+void rtc_timestamp() {
+	struct rtc_date rd;
+	rtc_get_clock(&rd);
+	rtc_print(&rd);
+}
+
+/**
+ *  Set Hardware Clock
+ */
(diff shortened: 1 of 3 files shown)


src/driver/rtc.c
```diff
@@ -0,0 +1,129 @@
...
+		}
+	}
+	
+	rtc_timestamp();
+}
+
+/**
+ *  Format RTC in a standard format
+ */
+static inline
```
bkleiner, line 73: please use `clang-format` to format these files, a config file is found in the root of the repository.
integrations from common editors are available.

bkleiner (changes requested):

src/driver/rtc.c
```diff
@@ -0,0 +1,129 @@
...
+				struct rtc_date rd;
+				sscanf(buffer,
+				       "%04d-%02d-%02dT%02d:%02d:%02d",
+					   &rd.year,
+					   &rd.month,
+					   &rd.day,
+					   &rd.hour,
+					   &rd.min,
+					   &rd.sec);
+			    rtc_set_clock(&rd);
```
bkleiner, line 61: could we have it skip this check if the current time is > the read time? would make testing less cumbersome as the file can stay on the card.

src/driver/rtc.h
```diff
@@ -0,0 +1,18 @@
+#ifndef _RTC_H
+#define _RTC_H
+
+struct rtc_date {
```
bkleiner, line 4: we are using the `typedef struct ...` pattern elsewhere. any reason we can not employ it here?

src/driver/rtc.c
SumolX: I was using notepad++, I will refactor.

src/driver/rtc.c
SumolX: I was thinking the same thing, I will update it

src/driver/rtc.h
SumolX: We could but I was keeping the implementation similar to rtc_time for consistency.  If you prefer typedef I can change it.

bkleiner: ok, yeah, i can see that. fair enough.

robhaswell:
Equivalent battery for UK: https://amzn.eu/d/7Jvjq36
Aliexpress: https://www.aliexpress.com/item/1005002584618945.html

src/driver/rtc.c
```diff
@@ -0,0 +1,141 @@
...
+    if (fd != -1) {
+        struct rtc_time rt;
+        if (ioctl(fd, RTC_RD_TIME, &rt) == 0) {
+            rt2rd(&rt, rd);
+        } else {
+            LOGI("ioctl(%d,RTC_RD_TIME,rt) failed with errno(%d)\n", errno);
+        }
+        close(fd);
+    } else {
+        LOGI("Failed to open(/dev/rtc, O_RDONLY)\n");
```
bkleiner, line 139: dito

src/driver/rtc.c
```diff
@@ -0,0 +1,141 @@
...
+    int fd = open("/dev/rtc", O_WRONLY);
+    if (fd != -1) {
+        struct rtc_time rt;
+        rd2rt(rd, &rt);
+        if (ioctl(fd, RTC_SET_TIME, &rt) != 0) {
+            LOGI("ioctl(%d,RTC_SET_TIME,rt) failed with errno(%d)\n", errno);
+        }
+        close(fd);
+    } else {
+        LOGI("Failed to open(/dev/rtc, O_WRONLY)\n");
```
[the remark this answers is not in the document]
SumolX: Will do.

bkleiner (approved):

bkleiner:
LGTM now, let me see if i can give it a quick test this weekend, i should have one of these cmos batteries somewhere.

SumolX:
Uploaded a new rtc.txt file which specifies a new format for setting the hardware clock.  The goal is to outline a filesystem safe format which can be used for reading/writing names and/or contents to/from files.

2013-03-14T09:31:30 -> 2013-03-14T09_31_30

src/driver/rtc.c
```diff
@@ -0,0 +1,266 @@
...
+    rtc_get_clock(&rd_now);
+
+    // Has time has accumulated since the
+    // the installation of the battery?
+    g_rtc_has_battery =
+        !(rd_now.year == 1970 &&
+          rd_now.month == 1 &&
+          rd_now.day == 1 &&
+          rd_now.hour == 0 &&
+          rd_now.min == 0);
```
bkleiner, line 137: this check currently produces false positives if the application is launched a > min after the goggles have booted.
shouldn't be a problem in regular use, but occurs quite frequently during development.
i propose we re-write to simply `g_rtc_has_battery = rd_now.year > 1970;`

src/driver/rtc.h
```diff
@@ -0,0 +1,72 @@
+#ifndef _RTC_H
+#define _RTC_H
```
bkleiner, line 2: we have a recent enough gcc, lets please go with a `#pragma once`

bkleiner (commented):
Running on my goggles _without_ a battery and my proposed change:

no rtc.txt
```
[INFO][rtc.c:rtc_init:135] rtc_init has NOT detected a battery
[INFO][rtc.c:rtc_print:83] 1970-01-01T00_13_49
```

with rtc.txt
```
[INFO][rtc.c:rtc_init:135] rtc_init has NOT detected a battery
[INFO][rtc.c:rtc_init:166] rtc_init is updating clock to a later date
[INFO][rtc.c:rtc_print:83] 2023-03-15T21_51_50
```

afterwards date returns the correct (GMT+0) date & time
```
# date
Wed Mar 15 21:53:12 GMT 2023
```

Code looks lovely!

src/driver/rtc.h
SumolX: Excellent!  I will make the change.

src/driver/rtc.c
SumolX: Yeah there really is no good way to check if the battery was installed.... especially after the initial installation of the battery.  I will make this change to simplify it.

PR #201, opened 2023-03-27
Allows for enter to override initial entry index.

SumolX:
If an enter callback is defined, we can now specify the initial entry point without defaulting to the first index.


SumolX:
These changes are now included in pull request https://github.com/hd-zero/hdzero-goggle/pull/206

PR #202, opened 2023-03-28
RTC init file date evaluation was invalid

SumolX:
Previous rtc.txt date check would update clock when the seconds/minutes/hours was greater than the current time.  We now take into consideration if the year/month/day is also greater than or equal to the current clock date.  rtc.txt can now be safely stored on the SD Card without causing unwanted updates to hardware clock.


SumolX:
This pull request has been superseded by pull request https://github.com/hd-zero/hdzero-goggle/pull/206

PR #204, opened 2023-04-04
merge with svn v412

ligenxxxx:
Changes:
- New Linux image with WIFI support
- Added uboot emergency update (No more current tedious PhoenixCard for debrick process)
- Improved analog decoder performance (RapidFire mode2 tested, should be no more vertical roll, black lines, etc)
- Fixed the bug causing REC solid on while not recording
- Fixed image offset for 1080p HDMI in
- Fixed bug of randomly entering into Focus chat during bootup
- Shorten boot time 
- Added HDZero 1080p30 mode  (VTX update is needed)
- Added WIFI networking streaming (Expansion V2 module is needed)
- Added soft switch for power on/off analog receiver in expansion module (V1 and V2). Available for Batch 2 goggles only.
- Added compatibility for batch 2 goggles (Have to change some parts, i.e. voltage ADC, and TP2825L for avialbility)
- Misc ( I have bad memory now)


chaotixtech:
Excellent work, this is a major update. 
I am curious what hardware difference there is from batch 1 to batch 2 that means the soft power switching of the module bay is not possible on batch 1. Controlling power to the bay is something I am interested in being able to do.

src/core/osd.c
```diff
@@ -550,21 +690,28 @@ void osd_signal_update() {
...
+        // display osd
         for (int i = 0; i < HD_VMAX; i++) {
             for (int j = 0; j < HD_HMAX; j++) {
-                uint16_t ch = osd_buf[i][j];
-                if (ch == 0x20)
-                    ch = elrs_osd[i][j];
-                if (ch != osd_buf_shadow[i][j]) {
-                    osd_buf_shadow[i][j] = ch;
+                if (osd_buf[i][j] != osd_buf_shadow[i][j]) {
+                    osd_buf_shadow[i][j] = osd_buf[i][j];
```
pkendall64, line 710: This change should be left as it was in the open source repo otherwise the ELRS backpack cannot add elements to the OSD.

ligenxxxx: This PR is just for synchronizing SVN and Git repositories. If there is a bug, we can submit another PR after this.

PR #213, opened 2023-04-10
Fix osd font loading

ligenxxxx:
- Reload osd font when video resolution(1080P<->720P) changed.
- Fixed the problem that osd_shadow cannot be fully cleared when switching osd font.

### DIFF
--- a/src/core/msp_displayport.c
+++ b/src/core/msp_displayport.c
@@ -185,8 +185,8 @@ void parser_rx(uint8_t function, uint8_t index, uint8_t *rx_buf) {
         parser_osd(index, rx_buf);
 }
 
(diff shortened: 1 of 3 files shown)


src/core/msp_displayport.c
```diff
@@ -233,7 +233,8 @@ void camTypeDetect(uint8_t rData) {
     }
     if (cur_cam == last_cam)
         CAM_MODE = cur_cam;
-
+    else if (cur_cam == VR_1080P30 || cur_cam == VR_1080P30)
```
pkendall64, line 236: Should this if be 
```cpp
else if (cur_cam == VR_1080P30 || last_cam == VR_1080P30)
```
so it switches back to the 720p font as well?

PR #214, opened 2023-04-10
Optimized iic open/close to a single open during initialization

SumolX:
Noticed we were hammering the iic open/close routines for every read/write operation when running the Emulator Build.  Modified code to open the ports during initialization and reference the file handles during read/write operations.  We will log a onetime error if one is detected.  Test goggles and emulator build to ensure no issues were introduced.

### DIFF
--- a/src/driver/i2c.c
+++ b/src/driver/i2c.c
@@ -16,14 +16,17 @@
 
 pthread_mutex_t i2c_mutex;
 
-int g_iic_fds[IIC_PORTS];
-static int g_iic_report_once[IIC_PORTS] = {0};
-static char *iic_ports[IIC_PORTS] = {
+#define IIC_PORTS 4
+
+static char *IIC_DEVS[IIC_PORTS] = {
     "/dev/i2c-0",
     "/dev/i2c-1",
     "/dev/i2c-2",
     "/dev/i2c-3",
 };
+
+int g_iic_fds[IIC_PORTS];
+static int g_iic_report_once[IIC_PORTS] = {0};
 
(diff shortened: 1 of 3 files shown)


src/driver/i2c.c
```diff
@@ -16,28 +16,41 @@
...
 
-static int iic_open(char *port) {
-    // int i2c_fd = open ("/dev/i2c-0", O_RDONLY);
-    int i2c_fd = open(port, O_RDONLY);
-    if (i2c_fd < 0) {
-        LOGE("iic_open %s failed [1=Right,2=Main,3=Left]", port);
+int g_iic_fds[IIC_PORTS];
+static int g_iic_report_once[IIC_PORTS] = {0};
+
+int iic_is_port_ready(int port) {
```
bkleiner, line 31: `stdbool.h` has been standardized for while :D lets please go with a bool.

src/driver/i2c.c
```diff
@@ -16,28 +16,41 @@
...
 }
 
-static void iic_close(int i2c_fd) {
-    close(i2c_fd);
+void iic_init() {
+    pthread_mutex_init(&i2c_mutex, NULL);
+    int i;
+
+    // Offset starts with 1 as it is not referenced thus far.
+    for (i = 1; i < IIC_PORTS; ++i) {
```
bkleiner, line 50: not a ton of reasons to pre-declare variables on modern compilers. lets please move iterator into the for statement.

src/driver/i2c.c
SumolX: Will do!

PR #217, opened 2023-04-12
Softswitch only available on Batch 2 and greater

SumolX:
Added hw_revision_t to common.
Restructured startup behavior to ensure hardware comms was available prior to deriving revision type.
Batch 1 users will not be shown softswitch option in power as it is only available on newer hardware.

Tested on Batch 1 goggle and Emulator.  Will need feedback from Batch 2 goggle users prior to approving pull request.

### DIFF
--- a/src/core/common.c
+++ b/src/core/common.c
@@ -15,7 +15,18 @@
 pthread_mutex_t lvgl_mutex;
 atomic_int g_key = 0;
 atomic_int g_init_done = 0; // 0= init not done, 1= done, -1= dial/up/down pressed
+static hw_revision_t g_hw_revsion = HW_REV_UNKNOWN;
 ///////////////////////////////////////////////////////////////////////////////
+
+void setHwRevision(hw_revision_t revision) {
+    if (revision > g_hw_revsion) {
+        g_hw_revsion = revision;
+    }
+}
+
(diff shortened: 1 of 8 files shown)


src/core/common.c
```diff
@@ -15,8 +15,19 @@
...
+        g_hw_revsion = revision;
+    }
+}
+
+hw_revision_t getHwRevision() {
+    return g_hw_revsion;
+}
+
 uint8_t slow_key(left_dial_t key, uint8_t *state, uint8_t *cnt) {
     if ((key == LEFT_DAIL_CLICK) || (key == LEFT_DAIL_LONGPRESS)) {
```
Zuldan, line 32: LEFT_DAIL_CLICK should be LEFT_DIAL_CLICK

SumolX: I agree spelling should be fixed but It should be handled in a different t pull request.

ligenxxxx:
I did a test in batch1 and found no problems.

SumolX:
> I did a test in batch1 and found no problems.

I was just notified that my batch 2 is shipped.  I will be able to test this change once it's delivered.

ligenxxxx:
Batch2 is OK.

PR #221, opened 2023-04-16
Dropbear a SSH/SCP server allowing for remote access via WiFi.

SumolX:
Provides the ability to SSH/SCP into goggles via WiFi.  Could be used for both debugging and transferring files remotely via SCP/WinSCP.

Eventually this service will be relocated to the OS distribution once we have the ability to update the OS distribution.  For now we will distribute this service as part of the application binary archive.

Note: Host key file will be automatically generated upon the initial installation and each user will have their own private host key.

### DIFF
--- a/ext/dropbear/CMakeLists.txt
+++ b/ext/dropbear/CMakeLists.txt
@@ -1,0 +1,65 @@
+cmake_minimum_required(VERSION 3.14)
+project (dropbear C)
+
+if(DEFINED CMAKE_TOOLCHAIN_FILE)
+	set(CMAKE_C_FLAGS "-mfpu=neon -mfloat-abi=hard")
+endif()
+
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-unused-function -Wno-unused-variable -ffunction-sections -fdata-sections -Wl,-gc-sections -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64")
+set(CMAKE_C_FLAGS_DEBUG "-g")
+set(CMAKE_C_FLAGS_RELEASE "-O3 -DNDEBUG")
+
+if(DEFINED CMAKE_TOOLCHAIN_FILE)
+	get_filename_component(COMPILER_NAME ${CMAKE_C_COMPILER} NAME)
+	get_filename_component(COMPILER_PATH ${CMAKE_C_COMPILER} PATH)
+	string(REGEX REPLACE "(.*)-gcc" "\\1" TARGET_PLATFORM "${COMPILER_NAME}")
+	set(TARGET_OPTIONS
+		"CFLAGS=${CMAKE_C_FLAGS}"
+		"CC=${COMPILER_PATH}/${TARGET_PLATFORM}-gcc"
+		"AR=${COMPILER_PATH}/${TARGET_PLATFORM}-ar"
+		"LD=${COMPILER_PATH}/${TARGET_PLATFORM}-ld"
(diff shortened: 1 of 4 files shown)


mkapp/app/script/rc.sh
```diff
@@ -75,6 +75,9 @@ usleep 2000
 #set Microphone Bias Control Register
 aww 0x050967c0 0x110e6100
 
+#dropbear process
+/mnt/app/script/dropbear_start.sh
```
bkleiner, line 79: lets detach the process here via `&` so blocking portions of the dropbear init (eg. keygen) do not hold up the boot process.

ext/dropbear/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+		"RANLIB=${COMPILER_PATH}/${TARGET_PLATFORM}-ranlib"
+		"NM=${COMPILER_PATH}/${TARGET_PLATFORM}-nm"
+		"AS=${COMPILER_PATH}/${TARGET_PLATFORM}-as"
+		"OBJDUMP=${COMPILER_PATH}/${TARGET_PLATFORM}-objdump"
+		"OBJCOPY=${COMPILER_PATH}/${TARGET_PLATFORM}-objcopy"
+		"STRIP=${COMPILER_PATH}/${TARGET_PLATFORM}-strip"
+		"STRINGS=${COMPILER_PATH}/${TARGET_PLATFORM}-strings"
+		"SIZE=${COMPILER_PATH}/${TARGET_PLATFORM}-size"
+	)
+	set(ENV ${TARGET_OPTIONS})
```
bkleiner, line 30: do we know if this is scoped in any way? or is this manipulating the env for the lifespan of the cmake process?

bkleiner (commented):
Frankly, i am still skeptical about this approach. Even for a interims solution it seems like a bad idea to me.
One particular concern is that it will make the transition to a rootfs-shipped variant harder, as port 22 will already be in use.

mkapp/app/script/rc.sh
SumolX: Agree will make the change.

ext/dropbear/CMakeLists.txt
SumolX: The later... performed the following tests:
Terminal 1: ran setup.sh and confirmed environment did not contain those environment variables.
Terminal 2: ran make -j from build directory and confirmed the environment variables were not present before or after.

SumolX:
> Frankly, i am still skeptical about this approach. Even for a interims solution it seems like a bad idea to me. One particular concern is that it will make the transition to a rootfs-shipped variant harder, as port 22 will already be in use.

If we ship a rootfs variant which has openssh built in it will create a conflict, but we can avoid this by specifying an alternate port when starting up dropbear. 10022 for example.  However, I believe we can avoid this all together by deprecating dropbear once a new rootfs-version is ready for release which would contain support for openssh.

Which leads me to the followup question, if we release a new rootfs variant, will we require a minimum version of the goggle application compatibility?  If so, then the Goggle Application will have deprecated dropbear and we can avoide incompatibility issues that would have occurred.

Hence, rootfs would require a minimum version of X.Y.Z for the goggle application which would not include dropbear.

The alternative solution is we don't use the goggle application to install dropbear, but this would require HDZ to officially update the HDZG_OS.bin to include this functionality instead and any potential maintenance or future updates can be isolated to the OS distribution vs having a convoluted application distribution which updates the installed OS.

SumolX:
Closing out this PR, Proposing an alternative method which will allow us to support various 3rd party services.

DavHau:
Why is the OS separate `.bin` to begin with?
Why not change the architecture to have one file including everything and flash everything at once?
This design choice seems to hold back development.

PR #224, opened 2023-04-19
Embedded OSD Glyphs optional SD CARD Location

shannonbaker:
Glyphs on SD card to be used, should exist as: /resource/OSD/GOGGLE and /resource/OSD/GOGGLE/FHD for 720p and 1080p respectively.  Kudos to bkleiner (Hanfer) for a lot of the code base for this!

### DIFF
--- a/src/core/osd.c
+++ b/src/core/osd.c
@@ -44,9 +44,8 @@ extern pthread_mutex_t lvgl_mutex;
 
 int gif_cnt;
 
-
-//Use SDCARD for Embedded Glyph if the glyph exists otherwise use goggle FS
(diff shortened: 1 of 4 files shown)


src/core/osd.c
```diff
@@ -61,10 +95,14 @@ void osd_llock_show(bool bShow) {
         return;
     }
 
-    if (g_latency_locked)
+    if (g_latency_locked){
```
bkleiner, line 98: couple of code-formatting errors, please use vscode's `format document` function.
that will format the code according to the clang-format file.

shannonbaker: Done with new commit.

PR #226, opened 2023-04-20
Add extended and transparent fonts for Kiss Ultra

KissUltra:
Add Extended and Transparent font for Kiss ULTRA


bkleiner:
does this deprecate the existing font at https://github.com/hd-zero/hdzero-goggle/blob/main/mkapp/app/resource/OSD/FC/UTRL_000.bmp ?

KissUltra:
yes, its new version.

KissUltra:
It has 2 pages and transparency :)

bkleiner:
do we need a transition period or can the old font be removed immeditatily?

bkleiner:
for consistency i do recommend that the font is submitted to https://github.com/hd-zero/hdzero-osd-font-library as well. that will allow VRX users to download it and allow goggles users to use the font without updating their firmware first.

KissUltra:
I know, but since its for ULTRA only, i keep all fonts on ULTRA wiki. Its NOT compatible with BF, so i rather have it in same place. ;) I had grace period of 5 releases to let pilots use SD card fonts, looks like now everyone updated - time to commit to HD systems fw.

bkleiner:
your call on how you want to handle distribution, but is that a yes on the immediate removal of the old font?

KissUltra:
Its a replacement font. same name, isnt it? Yes, new one should go in place of old one. Ligen already made me test build so we will test it tonight on googles.

bkleiner:
no, they have slightly different naming: https://github.com/hd-zero/hdzero-goggle/tree/main/mkapp/app/resource/OSD/FC
`UTRL_000.bmp` always seemed like a typo to me, but i dont have any Kiss gear to verify.

KissUltra:
CRAP!!!!!

KissUltra:
So, new file stays, old file can be deleted.

KissUltra:
You are the best. Thank you!

PR #222, opened 2023-04-17
Add new settings page and element positioning preview for goggle OSD

Nikolas-S:
**WARNING: The way settings are saved was modified on the main branch recently. If you test this PR and change the visibility of OSD elements, you will have to reset your settings using the reset button on the Firmware page just *before* going back to 9.0.12, otherwise your OSD elements may be invisible in 9.0.12.**

Created a new settings page for the goggle OSD.
Moved the osd mode setting from image settings to the new osd settings.

The new page allows for individual positioning and visibility settings for every goggle osd element.
The element position can be set separately for 4x3 and 16x9 modes, visibility is shared between modes.
The element position is set using x and y sliders. I think this is ok for now but the goal is to have the elements show and update live, similar to the image settings. (don't have my goggles yet and the osd seemed wonky on the emulator)

osd_page

I also added acceleration when moving the sliders to make positioning less tedious while not losing precision, would love some feedback on how it feels on real goggles.
If anyone has ideas for improvements, let me know.

For the devs: 
I added a new app state APP_STATE_SUBMENU_ITEM_FOCUSED.
When in that state, a pages on_roller is called but without automatically changing the submenu item selection, allowing the page to correctly handle input for items like dropdown menus and sliders.
I propose we use the new state for future changes, seems cleaner than separate app states for every element or other workarounds.

PR #249, opened 2023-05-19
Add button to reset all settings

### LINKED ISSUE #236: Feature: Add a way to reset all settings
As @pitts-mo discovered while testing pr #222 , currently, when updating the goggles, the old config file is always retained. 
When changes are made to the way the goggle handles settings, like with this recently merged pr #196 , retaining the old config file may have unintended consequences.

If I understand it correctly, a backup copy of the existing config file is made by [update_goggle.sh](https://github.com/hd-zero/hdzero-goggle/blob/a29cfcd692102f0d41329b41207d367830f22cae/mkapp/app/script/update_goggle.sh#L120) during the update process.
During startup, the function [settings_load](https://github.com/hd-zero/hdzero-goggle/blob/a29cfcd692102f0d41329b41207d367830f22cae/src/core/settings.c#L183) in settings.c then checks for the existence of that backup and, if it exists, replaces the actual config file with the backup and then deletes the backup.

Before the next release we should at least add a convenient way for the user to reset all settings, and possibly a way to automatically perform a settings reset during the update process if needed.

Nikolas-S:
Added a button to reset all settings to the Firmware page.
The button deletes the ini settings file, creates a new (empty) one, and writes the current version of the settings file to it.
Also added a new APP_STATE that blocks all user input, for example for when the user has to repower the goggles.
During startup, the version of the settings file is checked against the expected version and, if they're not the same, the settings are reset. This could be expanded to migration in the future.
With this it should also be possible to go back to older versions by resetting the settings right before updating.

Closes #236 

reset_btn

reboot_reset

Also added the new popup to the goggle update process, blocking input and prompting the user to repower.
reboot_update


Nikolas-S:
Would appreciate it if someone could give this a test on their goggles, don't have mine on me atm.

Nikolas-S:
I created a test [branch](https://github.com/Nikolas-S/hdzero-goggle/tree/test/setting_reset_osd_element_pos) on my fork, based on the current main, that has #222 and this pr merged.
@pitts-mo If you have the time, would you mind testing the branch and after changing the osd element positions, use the reset button just before going back to 9.0.12? https://github.com/Nikolas-S/hdzero-goggle/actions/runs/5043798284
Hopefully the osd elements will be ok when going back to 9.0.12.

pitts-mo:
This was already my first priority when my goggles return from optic replacement. ETA within next couple days? I will verify 9.0.12 is still installed and build this PR for SD card. :-) 
Thank you,
-p

ligenxxxx:
It would be nice if the edge of the pop-up window is red so that we can keep the UI style consistent.

Nikolas-S:
Was going with green initially to indicate ok/warn/error but trying it with red, I think you're right, fits better.
border_red_text_green

But I do think we should keep using the title color to indicate outcome, examples for warn/error:
border_red_text_yellow
border_red_text_red

pitts-mo:
@Nikolas-S 
So far the settings reset appears to work as expected. I performed a reset and booted to 9.0.12 and was able to regain the old OSD functionality. I performed another reset test after returning to this PR without issue.

However, I notice a three odd characters keep getting appended onto the APP section of the Current Version display. The characters appear random for each boot session and persist until power cycle.

Nikolas-S:
@pitts-mo 
Thank you for testing!
I found a possible issue with how the version is read from file, when you have time, please let me know if the issue persists.
If it does, please add an example of what exactly your version says.

pitts-mo:
Resolved.
I no longer observe erroneous characters appended to the APP version or anywhere else within the Current Version string.
Thank you,
-p

PR #256, opened 2023-05-25
New Virtual Keyboard, WiFi Module Page and ELRS Backpack Page

SumolX:
Introduces a virtual keyboard used to configure text fields within the wifi module configuration page.
Refactors Page Connections and breaks them apart into their respective pages (ELRS Backpack, WiFi Module)
Deprecates ssid.txt and allows for complete configuration of wifi module via User Interface.
WiFi Page contains all required options to manage the WiFi Module.

All scripts are auto-generated from the WiFi Module Page upon creation and modification.  The only script referenced is the wlan_stop.sh script which was renamed from wlan_stop_ap.sh as it applies to both modes of operation.

New Virtual Keyboard Interface:
* Keyboard for entering input text fields.
  * Left Click Enables/Disables keyboard when text input is highlighted
  * Scroll Wheel highlights a key
  * Right Short Click confirms/presses highlighted key
  * Right Long Click clears text entry.

New WiFi Module Interface:
* Paging through WiFI configurations (Basic/Advanced).
  * Basic Configuration:
    * Enable/Disable WiFi Module
    * Specify mode of operation: Host (Access Point) or Client (Join Network)
    * Unique SSID and Password for both Host and Client modes of operation.
    * Apply Settings
    * Return Back to Main Menu
  * Advanced Configuration:
    * DHCP Client enable/disable (Only Applies to Client Mode)
    * Ip Address - Applies to both modes
    * Netmask - Applies to both modes
    * Gateway - Applies to both Modes
    * DNS Server - Applies to both Modes

For a comprehensive list of features I have uploaded 2 videos (Demo 1 and 2).
Demo 1:
https://youtu.be/wiCCv1Q3zlM

Demo 2:
https://youtu.be/krB9vMg_SVE


ligenxxxx:
When DHCP on is to hide address, netmask and other options, what do you think?

SumolX:
> When DHCP on is to hide address, netmask and other options, what do you think?

Those fields are also used for configuring Host (Access Point) mode and for Client (Join Network) mode.  When DHCP is on, the request includes the  IP Address, if no other device on the network is using that ip address then it will be used, otherwise a unused address will be assigned.  It's a nice option because you may have a configured range where you still want to use dhcp but be able to specify an address for consistency.

PR #258, opened 2023-05-29
Reduce  Dial Sensitivity

Knifa:
Reduces the left dial sensitivity to require two clicks instead of one, with a 1s timeout.

At least in my little group of HDZero owners, we all think the dial is too sensitive simply because it requires very little force to turn it. This makes trying to get to the right menu item or channel trickier than it should be. Based on that, I really think this should just be the default! 

It looks like there were provisions for this already with `slow_key` but it is unused. I've removed this and related defines, etc. that are not used.

Happy to add configuration if folk think otherwise though --- though even then I'd opt for this to be the default and the configuration is to make it *more* sensitive rather than less.


Knifa:
I take it back, maybe making it the default is too strong. 😅 I do prefer it though so I'll have a look at making it configurable.

evilC:
> I really think this should just be the default!

Have you tried scrolling through the playback menu with 300 clips in it? 🤣 
TBH for that, we kind of need the opposite.
Maybe some logic that detects that you are doing LOTS of scrolling and starts to scroll in increments of 10...

ligenxxxx:
For me, `#define DIAL_SENSITIVITY 2` is awkward to operate.
Please add configuration, but set `DIAL_SENSITIVITY` to 1 by default.

Knifa:
> For me, `#define DIAL_SENSITIVITY 2` is awkward to operate. Please add configuration, but set `DIAL_SENSITIVITY` to 1 by default.

Bumped it down now.

Think this could be merged if you like, can add UI config, etc. for it later.

PR #260, opened 2023-05-29
Offload beep to new thread, make audible.

Knifa:
This PR moves beep GPIO control to a separate thread so that beeps don't block and so can be made long enough to be audible. 😄 Calling parts of the system don't need to worry about it waiting anymore.

- The previous beep (via `beep()`) is only 500us (!) so just makes a little tap sound. The default beep is now 50ms, with a longer 200ms beep available for future use.
- `beep_dur` now takes milliseconds, rather than microseconds.

### DIFF
--- a/src/core/elrs.c
+++ b/src/core/elrs.c
@@ -26,7 +26,7 @@
 #include "core/osd.h"
 #include "core/settings.h"
 #include "driver/dm5680.h"
-#include "driver/gpio.h"
+#include "driver/beep.h"
 #include "driver/hardware.h"
 #include "driver/rtc.h"
 #include "driver/uart.h"
--- a/src/core/elrs.c
+++ b/src/core/elrs.c
@@ -295,7 +295,7 @@ void msp_process_packet() {
             }
         } break;
         case MSP_SET_BUZZER:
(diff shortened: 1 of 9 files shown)


Knifa:
Paired with #258, for example! Used `beep_dur(1)` on this for some satisfying little taps. No blocking!
https://github.com/hd-zero/hdzero-goggle/assets/622619/1e2b63f5-648d-49c5-a1de-3428ae15046f

Knifa:
Moved some stuff around so the thread only needs to lock the mutex once.

SumolX:
I've never noticed that there was supposed to be an audible sound outside the clicking of the scroll wheel.  Nice job!

ligenxxxx:
Not everyone likes the buzzer beeping when scrolling the scroll wheel. How should we meet different user needs?

Knifa:
> Not everyone likes the buzzer beeping when scrolling the scroll wheel. How should we meet different user needs?

This isn't included in the PR. Just an example of how the beep can work now that it doesn't block the thread.

ligenxxxx:
It is wise to implement the buzzer in a separate thread.
Also, I suggest moving beep related functionality to beep.c. Because I have extra ideas to enrich the interface and control of beep, so that it can even play some simple 8bit music.

Issue #267, opened 2023-06-02
v9.1.0-rc1 Go Sleep! -restore analog state -resume improvements


pitts-mo:
The AnalogRX Power ON state is not preserved after utilizing Go Sleep! function.

To reproduce:
-Set AnalogRX Power ON within Power menu.
-enable live Analog view and verify module is powered and working.
-return to menu and select Go Sleep! option. (goggle displays and analog module are turned off -this is desirable)
-when ready to press a key to resume.
-goggle displays will turn on with user returned to menu but the analog bay is left off (will show AnalogRX Power=ON in menu until power cycle)

Secondary items:
I believe it is more desirable to return the user to the most recent Live view instead of the menu after resuming from sleep.
It may be beneficial to work issue toward resolution of #230 while working this issue.

pitts-mo:
Note: I have batch 2 goggle. I suspect the analog module will remain powered during Go Sleep! on the batch 1 goggle.

pitts-mo:
The primary concern for this ticket was resolved by PR#268 included within 9.1.0-rc2. Closing issue.

Issue #230, opened 2023-04-24
Feature Request: add AnalogRX Power option "Auto"


pitts-mo:
Please add "Auto" setting option to "AnalogRX Power" within the power settings page of UI **on >=V2 batch goggles**.

Make "Auto" the default for this option.

If "Auto" 
check/set analog module power is ON when:
-goggle is booting and analog is the current mode ~~or WiFi / other analog module bay feature is enabled~~
-mode is changed to analog ~~or WiFi / other analog module bay feature changes to enabled~~
else
check/set analog module power is OFF when:
-goggle is booting and analog is NOT the current mode ~~and WiFi / other analog module bay feature is NOT enabled~~
-mode is changed to NOT analog ~~and all WiFi / other analog module bay features change to NOT enabled~~

**2023-07-02 update:
The analog module v2 WiFi is powered independently and continues to function even if the analog module soft power switch within the v2 goggle has been set OFF.**

pitts-mo:
$ grep -r -i power_ana
mkapp/app/setting.ini:power_ana_rx = 1
src/ui/page_power.c:    ROW_POWER_ANA,
src/ui/page_power.c:static btn_group_t btn_group_power_ana;
src/ui/page_power.c:        create_btn_group_item(&btn_group_power_ana, cont, 2, "AnalogRX Power", "On", "Off", "", "", ROW_POWER_ANA);
src/ui/page_power.c:    btn_group_set_sel(&btn_group_power_ana, g_setting.power.power_ana);
src/ui/page_power.c:    case ROW_POWER_ANA:
src/ui/page_power.c:            btn_group_toggle_sel(&btn_group_power_ana);
src/ui/page_power.c:            g_setting.power.power_ana = btn_group_get_sel(&btn_group_power_ana);
src/ui/page_power.c:            ini_putl("power", "power_ana_rx", g_setting.power.power_ana, SETTING_INI);
src/ui/page_power.c:            DM5680_Power_AnalogModule(g_setting.power.power_ana);
src/core/main.c:    DM5680_Power_AnalogModule(g_setting.power.power_ana);
src/core/settings.c:        .power_ana = false,
src/core/settings.c:    g_setting.power.power_ana = ini_getl("power", "power_ana_rx", g_setting_defaults.power.power_ana, SETTING_INI);
src/core/settings.h:    bool power_ana;
src/driver/dm5680.h:void DM5680_Power_AnalogModule(uint8_t on);
[...]

Issue #223, opened 2023-04-18
GoSleep! UX improvements


robhaswell:
Even with the new update, the boot time is very high. This can lead to you holding up a race if you forget to turn them on.

It might be possible to reduce power consumption to the point where the units don't need to be powered down. I suggest that a mode could be activated (by way of some button press, perhaps both buttons at once) to enter "standby mode".

In standby mode, the HDZero radios and OLED as powered down, and there could be additional firmware actions which reduce power consumption of the MCU. Fans are disabled or run at minimum. Standby mode ends as soon as any button is pressed.

Standby mode could also be entered automatically after a period of no signal, similar to the current OLED-off function.

---

## Edit

Since this issue was created, a new function "GoSleep!" has been added to the menu. I've updated this issue to represent my suggestions for UX changes to GoSleep! See comment https://github.com/hd-zero/hdzero-goggle/issues/223#issuecomment-1598843309

pitts-mo:
@robhaswell a Go Sleep! function recently added and I have submit PR #268 with some improvements.

In your view of Standby mode... should the goggle attempt to immediately return to Live view on resume? or would you prefer it to return to the menu with "Go Sleep!" still highlighted?

Anyone else wish to weight in?
Thank you,
-p

pitts-mo:
As of 9.1.0-rc2 the Go Sleep! function appears to provide the basic hardware sleep functionality needed to resolve this issue. However, performance and usability are lacking.

Outstanding items I see are:
-make Go Sleep! function independent of menu.
-add logic for corded long left + right button press to actuate Go Sleep! function.

Sound correct?
If so maybe @robhaswell can update the title for this issue in a few days?

SAMUD:
I updated my googles to the lastet version and the GoSleep is cool. However some quick acess for this is needed.
Either both buttons or some kind of double click on the rec-button to send them to sleep.

Optional I think something like one beep every 30sec like on Orqa googles would be cool, so you are reminded that the googles are still powered on.

PR #268, opened 2023-06-03
Go Sleep! impovements -restore analog power state

pitts-mo:
Go Sleep! exit behavior improvement.

Batch 2 goggles will now:
-force analog power off instead of setting analog power off.
-use existing analog power setting to restore analog power state when exiting sleep.

~~All goggle batches will now:
-attempt to resume live view immediately upon exit.~~

This is a good enough resolution for #267.
However, #230 still needs to be addressed and I still see it desirable to unify all power and state setting into a single location.

### DIFF
--- a/src/ui/page_sleep.c
+++ b/src/ui/page_sleep.c
@@ -1,6 +1,7 @@
 #include "page_sleep.h"
 
 #include "core/common.hh"
+#include "core/app_state.h"
 #include "core/settings.h"
 #include "driver/dm5680.h"
 #include "driver/fans.h"
@@ -48,9 +49,7 @@ static void page_sleep_enter() {
 
     // Turn off Analog Receiver  -- Batch 2 goggles only
     if (getHwRevision() >= HW_REV_2) {
-        g_setting.power.power_ana = 1;
-        ini_putl("power", "power_ana_rx", g_setting.power.power_ana, SETTING_INI);
-        DM5680_Power_AnalogModule(g_setting.power.power_ana);
+        DM5680_Power_AnalogModule(1);
     }
 
     // Minimum fan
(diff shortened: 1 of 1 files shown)


ligenxxxx:
I don't think **return to live video after sleep exit** is a good idea.
I think where you enter sleep mode, you should return there after exiting.

pitts-mo:
I see the intent of this feature as saving power and reducing boot time... I have changed the usage hint and asked another steak holder.

"Click the Enter Button to go sleep.\nClick any button to exit sleep mode and return to Live view."

Please comment if anyone else has a preference for the post sleep behavior.
Thanks,
-p

pitts-mo:
@ligenxxxx I only see one related comment on #223. 

Do you still object to the temporary improvement of returning to Live view on resume from sleep?

... If you still object and it seems unlikely someone will be working the Go Sleep! function independent from the menu I would be happy to separate this into two PR so the fix to restore analog power state can be merged more quickly.

robhaswell:
I definitely think returning to live view is the right thing to do, but as an alternative, we could honour the user's preference on what to do on the goggle's first boot? There is an option of whether to show the goggle menu or last viewed live view. In a way, resuming from sleep is the same user experience as booting up, so honouring that preference would make sense.

pitts-mo:
As far as I can tell the way this PR resumes from sleep and results in the same as boot up.

Unless there is a setting combination that ignores scan and all source options allowing you to persist at the menu after booting?

ligenxxxx:
How about this:
`Click the Enter Button to go sleep.`
`Click any button to exit sleep mode and enter menu mode.`
`Long press any button to exit sleep mode and enter video mode.`

pitts-mo:
Attempting to preserve a UI experience where it would deviate from normal boot by changing the logic of user interaction makes even less sense than removing the attempt to make sleep resume in the same way as a fresh boot.

That said, the comment did prompt me to test and discover some app state behavior I am not able to quickly understand.... 

So I have removed the resume to live view / result of fresh boot behavior for a future fix.

PR #283, opened 2023-06-15
Send MSP to backpack when channel is changed for HDZero

cruwaller:


cruwaller:
First approach was a bit rude and decided to move it under ELRS menu.

cruwaller:
rebased on main.

HazzaHFPV:
So, how does this work, what does it do? I see the entry in the menu, but it doesnt do anything..or i am not understanding it.

cruwaller:
Functionality is on the way and this allows to set VTX channel from the goggles.

PR #284, opened 2023-06-15
Add option to set startup OSD visibility

### LINKED ISSUE #188: Goggle OSD Setting for Startup
I would like to see a software option toggle to disable/enable the goggle's OSD elements on bootup.

Currently the goggle's OSD elements are always enabled on every power up.  Even if they were turned off prior to the last power off.  I would like to see an option that allows the goggles to power on without any of the goggle's OSD elements being displayed unless called for via the manual button press as it is currently configured.

Thank you for your time and hard work!

Nikolas-S:
Added a new option to the OSD page to set the startup visibility of the goggle OSD.
Closes #188 

osd_startup_vis


cruwaller:
This is nice!

PR #286, opened 2023-06-18
WiFi Module: added 3rd Page to configure System Options

SumolX:
Some users have started to receive their wifi modules and are unable to ssh into goggles.  This is due to a OS root password bug which relates to /etc/passwd and the shadowing of the password not being setup properly.

New 3rd Page has been added which allows users to configure the root password and enable/disable SSH access (defaults to disabled).

Thus protecting their system from individuals who may perform nefarious deeds when logged into or connected to the same WiFi network as their goggles.

image

### DIFF
--- a/mkapp/app/script/wlan_stop.sh
+++ b/mkapp/app/script/wlan_stop.sh
@@ -4,6 +4,7 @@ sleep 1
 killall rtspLive
 killall hostapd
 killall udhcpd
+killall dropbear
 ifconfig wlan0 down
 rmmod xradio_wlan.ko
 rmmod xradio_core.ko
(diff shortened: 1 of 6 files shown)


SumolX:
Ignore the first commit as we now have a dedicated page which allows the user to set their own root password and enable/disable ssh access.

MrAlucardDante:
Just like we can set the wifi ssid and password with a text file on the sd card, it could be nice to setup ssh the same way too, for consistency. 

For instance:
```
[ssh]
password=my_pass
```

And it could work like this « if the [ssh] block is present, then ssh is enabled, otherwise it’s disabled »

SumolX:
> Just like we can set the wifi ssid and password with a text file on the sd card, it could be nice to setup ssh the same way too, for consistency.
> 
> For instance:
> 
> ```
> [ssh]
> password=my_pass
> ```
> 
> And it could work like this « if the [ssh] block is present, then ssh is enabled, otherwise it’s disabled »

The text file was deprecated in favor for the dedicated WiFi Module page.

PR #288, opened 2023-06-19
Add wraparound to playback menu (Fix #275)

### LINKED ISSUE #275: Playback menu lacks wrap-around
In the playback menu, if you scroll back when the first clip is selected, it should go to the last clip.  
In the case where you have lots of clips on the SD card, it takes aaaages to scroll forwards through to the first clip.  
This is especially annoying when using the Hot Clips feature, as these are placed at the start of the list.  
ie you do something cool, go into playback menu, mark `hdz_250` at HOT, and then if you want to review it, you currently need to scroll through 250 clips to get to it. If this feature were implemented, you could simply scroll back, and then you are at the first hot clip

evilC:
This code is COMPLETELY UNTESTED as I am unable to get building working on my (Windows 10) PC so far.
However, the code seems pretty straightforward (Although I am not a C coder, so could have completely mis-understood the code) so I thought it worth submitting a PR anyway

### DIFF
--- a/src/ui/page_playback.c
+++ b/src/ui/page_playback.c
@@ -367,7 +367,7 @@ void pb_key(uint8_t key) {
     done = false;
     switch (key) {
     case DIAL_KEY_UP: // up
-        if (media_db.cur_sel = (media_db.count - 1)) {
+        if (media_db.cur_sel == (media_db.count - 1)) {
             media_db.cur_sel = 0;
         } else {
             media_db.cur_sel++;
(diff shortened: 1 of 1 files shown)


src/ui/page_playback.c
```diff
@@ -367,17 +367,21 @@ void pb_key(uint8_t key) {
     done = false;
     switch (key) {
     case DIAL_KEY_UP: // up
-        if (media_db.cur_sel != (media_db.count - 1)) {
+        if (media_db.cur_sel = (media_db.count - 1)) {
```
pkendall64, line 370: That should be == otherwise it's an assignment!

evilC: Oh my, schoolboy error. Fixed

SumolX:
Tested works great!

PR #290, opened 2023-06-20
fix sdcard integrity check and format scripts

antony-weber:
Fix sdcard scripts to support correct sdcard partition.
Add compatibility with sdcards formatted without partition table by previous scripts.

### DIFF
--- a/mkapp/app/script/chkfixsd.sh
+++ b/mkapp/app/script/chkfixsd.sh
@@ -13,14 +13,19 @@ else
     echo "Umounting SD Card: FAILURE"
 fi
 
+BLKDEV=/dev/mmcblk0p1
+if [ ! -b "$BLKDEV" ]; then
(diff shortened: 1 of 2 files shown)


SumolX:
Nice.  You must have formatted the card outside of the goggles?  I will test on my goggles to ensure compatibility.

antony-weber:
New sdcards have partition table by default. Formatted by previous script have no partition table. Whole block device formated as one partition. After integrity check update only cards without partition works for me. Other fails check. Googles say "no sd card". 
I test this fix with 4 different cards:
- new card with default partition;
- card formated manually from my laptop with no patririon table (`mkfs.vfat -F 32 /dev/sda -n "HDZERO"`) ;
- card formatted by googles ( previous script);
- card from my prevoius googles with partition table;
All cards works fine.

PR #293, opened 2023-06-21
Reduce emulator CPU usage

Nikolas-S:
Changing the SDL user input detection from poll to wait significantly reduces the emulators CPU usage.
I don't know if there was a reason why poll was chosen initially but wait seems to work fine.


bkleiner (approved):

SumolX:
Tested looks good

cruwaller (approved):

PR #300, opened 2023-06-26
Enable settings.ini for emulator

cruwaller:


SumolX:
This is a nice modification, my directory structure is as follows in order to satisfy the emulator:

/mnt/app -> ~/hdzero-google/mkapp/app (symlink)
/mnt/extsd (created directory)

The downside is i always have to undo my settings.ini prior to committing changes.

src/ui/ui_style.c
```diff
@@ -32,18 +32,6 @@ int style_init(void) {
     lv_style_set_pad_right(&style_rootmenu, 0);
     lv_style_set_radius(&style_rootmenu, 0);
     lv_style_set_border_width(&style_rootmenu, 0);
-
```
SumolX, line 35: Why are these being removed?  Are they not referenced?

cruwaller: Duplicate code... Just cleanup

cruwaller:
> This is a nice modification, my directory structure is as follows in order to satisfy the emulator:
> 
> /mnt/app -> ~/hdzero-google/mkapp/app (symlink) /mnt/extsd (created directory)
> 
> The downside is i always have to undo my settings.ini prior to committing changes.

Yeah, I run emu from build folder so ini gets removed when rerun cmake.

SumolX:
> > This is a nice modification, my directory structure is as follows in order to satisfy the emulator:
> > /mnt/app -> ~/hdzero-google/mkapp/app (symlink) /mnt/extsd (created directory)
> > The downside is i always have to undo my settings.ini prior to committing changes.
> 
> Yeah, I run emu from build folder so ini gets removed when rerun cmake.

Nice tip!  Does the same apply for recorder settings as well? thats another one i have to revert changes too when running emulator.

cruwaller:
I haven't tested the recorder ini. Need to check that as well... But I guess it should be similar to this.

ligenxxxx:
- I modified the power parameters through emulator.
- Restart the emulator and check the parameters, the last modification worked, no problem.

PR #301, opened 2023-06-26
Harmonize video file naming to avoid mismatch

cruwaller:
This PR moves all video file naming definitions to one file.

There is still a mismatch where `recorder` uses different folder than UI video playback. Recorder is using config file which can be used to set video output into different folder than default `/mnt/extsd/movies/`.


ligenxxxx:
@cruwaller Please slove the conflicts.

cruwaller:
Thanks! I will fix.

cruwaller:
rebased to main.

PR #306, opened 2023-07-03
Enable formatting not mounted sdcards

antony-weber:
Add block device exist check instead of sdcard mounted check. Now googles can format sdcards with unsupported file system to FAT32. 
Large capacity sd cards has exFat file system by default.

By the way, it would be nice to add support for exFat and ext4 in future releases.


christhomas:
Yeah this is a pretty good solution. Check the device exists instead of can be mounted. Then cards inserted with the wrong filesystem can be reformatted. I like it.

PR #312, opened 2023-07-08
AnalogRX Power "Auto" -batch 2 goggle

pitts-mo:
This PR replaces the AnalogRX Power setting "Off" with "Auto" and adds a function to track/set analog module bay power only on state change. Currently this will beep on every commanded state change except initiation of Go Sleep! function.

WARNING!
This should read like a very familiar hangup with implementing a physical power switch. Adding this feature is important. However, creating any system to change the power state of a critical item introduces potential risk. So, This needs to be heavily tested prior to usage with remote equipment and commit of this PR into master.

Known issues while AnalogRX Power is set to "On":
-none this is similar to batch 1 goggle behavior

Known issues while AnalogRX Power is set to "Auto":
-Source page will report Expansion Module: Disconnected when Expansion Module is not the currently selected source. Just select Expansion Module to use it anyway.
-when using ELRS backpack without a Binding Phrase (traditional binding) in your analog bay you will loose your backpack binding by switching in/out of Expansion Module source within 30 seconds three times.

Potential future enhancement:
-Add "Off" back in as a distinct setting (On, Auto, Off).
-disable beep on state change.
-add priority settings for debug beeps.

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -124,7 +124,6 @@ static void device_init(void) {
     TP2825_Config(0, 0);
     DM5680_req_ver();
     fans_top_setspeed(g_setting.fans.top_speed);
-    DM5680_Power_AnalogModule(g_setting.power.power_ana);
 }
 
 void lvgl_init() {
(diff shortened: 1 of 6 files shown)


pitts-mo:
Ok, I have rebuilt from this PR and Auto AnalogRX Power function appears to resolve issue #230 .

Suggest running HDZGOGGLE from SD card and heavily test this is working as you expected before operating your remote equipment .
-p

PR #317, opened 2023-07-24
Prevent corrupt MP4 when powerloss occurs.

SumolX:
Moving forward we will utilize fragmented MP4.  Fragmented MP4 segments the mdat (Movie Data) to the mp4 file based on I-Frame detection.  This allows for the file to still be usable even though the last mdat was not written.  For more information I recommend this read as it explains the differences between Fragmented and Non-Fragmented:

https://stackoverflow.com/questions/35177797/what-exactly-is-fragmented-mp4fmp4-how-is-it-different-from-normal-mp4

### DIFF
--- a/src/record/ai2aenc.c
+++ b/src/record/ai2aenc.c
@@ -21,6 +21,7 @@
 #include <unistd.h>
 
 #include "ai2aenc.h"
+#include "record.h"
 
 static void ai2aenc_configAiAttr(Ai2Aenc_t *aa, AiParams_t *param) {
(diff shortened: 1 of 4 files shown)


SumolX:
@evilC has confirmed fragmented MP4 is compatible with Premiere Pro and Kdenlive.  However, I have converted this PR to a draft as it should not be merged into master until I have the Goggle playback working with this file format.

SumolX:
Closing out PR since the goggles built-in MP4 decoder does not support fragmented MP4 playback.  Video playback duration are not decoded properly.

PR #323, opened 2023-07-30
add OSD:Clock

tkusukawa:
Since the RTC worked, it would be nice to be able to display the clock on the OSD and record.
https://twitter.com/i/status/1685572874994872320


nurikk:
Terrific work mate! Is it possible to switch between full datetime and hh:mm formats?

SumolX:
The only issue I have is the movement of the AM/PM format while the seconds are iterating in the clock page.  I had broken them up to prevent this from occurring.  Otherwise its a very nice addition and perhaps the same technique should be applied to the HUD as well.  Keep date, time and format as 3 individual UI elements, thus allowing users to configure what they want to see.

tkusukawa:
OK. I'll try. Please wait a moment.

tkusukawa:
@nurikk , @SumolX 
how about this

SumolX:
I will test it out tonight.  Thank you for making the modifications.

SumolX:
I have 3 comments:

1. On the Clock Page the AM/PM/24 hour format indicator is shifting left and right while the seconds are incrementing.   Please restore the format back to an individual UI element as it was previously before.  Which in turn means the OSD should also have a UI element for format (AM/PM/24 hour).  The format indicator should appear static and not float left-to-right based on seconds incrementing, the selected font effects the presentation, thus having it as a separate UI element resolves this issue.

2. Since the goggles do not come with a RTC battery pre-installed, the default behavior should be not to show Date/Time/Format.

3.  The font used for DATE/TIME/FORMAT should be derived from flight controller and not be hardcoded... please refer to the existing examples in the code for consistency.

Overall I like it alot and can't wait to use it on my goggles.

tkusukawa:
@SumolX 

There is a message "This workflow requires approval from a maintainer." on This GitHub Page.

Thanks for pointing out the branch. I will do that from now on.

I considered calculating AM/PM position from the TIME position, but it would be too complicated. If we don't want to move the position by content of string, we should use a monospaced font.

I love hidden tricks.

I didn't like that font either, so I changed it back.

tkusukawa:
Thank you for the discord suggestion. However, I am not good at English, so the discussion here will help me.

Please forgive me for no response a few days as I will be going on a mountain climb.

Thank you for the font information, too. I'll consider both.

pitts-mo:
The linux date command is extremely easy to work with to trans-code or parse out offsets and segments of DateTime including 12/24 hour and AM/PM. I do remember seeing the date command mentioned in the busybox PR#314

Is date already included in the base OS? 
If so is the app able to accesses commands from the OS?

Look forward to this feature. I figure the user should be allowed to enable the individual DateTime segments they want and load their own fonts for sizing.

Thank you for working on this,
-p

SumolX:
@tkusukawa 

I've pushed up a PR which contains the more common font sizes, this will allow you test the various sizes in order to keep the mapping between FC and OSD presentation as it provides a cohesive user experience.

https://github.com/hd-zero/hdzero-goggle/pull/326

@pitts-mo

Personally I would use the C language calls in regards to display formatting... using the date command would result in an external system call then we would need to store the standard output to a file and finally parse the file.  The C language provides this functionality via `strftime`.

tkusukawa:
@SumolX 
I tried to display AM/PM in a separate frame, but I don't think it's good.
-> https://github.com/tkusukawa/hdzero-goggle/commits/separate_frame_ampm
Because I couldn't find a way to calculate the offset, so I had to embed a constant value.

Also, I haven't found a way to specify a font or make it a fixed width.

I think that moving AM/PM is unavoidable unless there is a way to display fixed-width characters.

SumolX:
I used most of your source to create a version which displays everything in a static form and the AM/PM Format indicator is bound to the Time element.

https://github.com/SumolX/hdzero-goggle/tree/osd_datetime

tkusukawa:
@SumolX 
Thank you for progressing!
I am satisfied enough if exist the function of clock OSD.
I hope other people will benefit from it.
And I can't take care the complicated code, so it would be helpful if your code are pulled.

SumolX:
@tkusukawa I will create a PR then and you can close this one out.

tkusukawa:
@SumolX 
Thank you very much!

PR #331, opened 2023-08-11
OSD Clock Added

SumolX:
In collaboration with @tkusukawa we have added the RTC/Clock to the OSD.  Refer to PR https://github.com/hd-zero/hdzero-goggle/pull/323

By default the clock will not be displayed as most users have not installed the RTC battery but for those that have they now have the option to enable it and position the Date and Time as they see fit.


tkusukawa:
The time was not updated in the normal OSD display.

I think that "osd_clock_show(g_setting.osd.is_visible);" is necessary 
around "src/core/osd.c l.565:   osd_battery_voltage_show(g_setting.osd.is_visible); ".

And, Shouldn't "src/core/osd.c l.552: osd_clock_show(g_setting.osd.is_visible);"
be "osd_clock_show(true);"?

SumolX:
> 

Not sure how I missed that... nice find and fixed!

PR #339, opened 2023-09-04
Let pages handle roller interaction

Master92:
As stated in [app_state.h:27](https://github.com/hd-zero/hdzero-goggle/blob/b1c7d90f0dc3db1aa27b2d7cb3bc7051a45df158/src/core/app_state.h#L27), "pages should set the on_roller callback and handle the input themselves [...]".

This PR updates the corresponding pages' compile units to do that and removes the enum entries from the app_state header. It should be purely cosmetical and shouldn't change any behavior.

Tested my changes with a set of batch 2 goggles and couldn't find any unexpected behavior - however another one testing it would be appreciated.

### DIFF
--- a/src/core/app_state.h
+++ b/src/core/app_state.h
@@ -23,10 +23,6 @@ typedef enum {
     APP_STATE_WIFI = 13,
 
     APP_STATE_USER_INPUT_DISABLED = 20,
-
-    // TODO pages should set the on_roller callback and handle the input themselves
-    // instead of creating an app state. (use APP_STATE_SUBMENU_ITEM_FOCUSED)
-    PAGE_ANGLE_SLIDE = 101,
 } app_state_t;
 
 extern app_state_t g_app_state;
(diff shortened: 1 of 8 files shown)


ligenxxxx:
@Master92 Please fix the bug.

Master92:
> @Master92 Please fix the bug.

I'd really like to. Please describe what the bug is.

ligenxxxx:
src/ui/page_headtracker.c [line 201]
This function contains app_state_push(APP_STATE_SUBMENU) and causes state confusion and app crash when exiting HT submenu.
When exiting the submenu, g_app_state is set to APP_STATE_MAINMENU in submenu_exit(). But app_state_push(APP_STATE_SUBMENU) overrides its state.
I guess you just want to reset some states. I think it is enough to reset these states when entering the submenu. There is no need to reset both when entering and exiting.

Master92:
Good catch!

I reproduced your bug when entering either the fans or the headtracker page and exiting them with the back button. Headtracker page made the app indeed crash while in fans page you simply wouldn't exit it.
This mis-behavior was caused by unconditionally calling the slider exiting methods. Those are in turn necessary when exiting the page directly with a long press of the left button while editing the sliders.

I now call those methods only when a slider has really been selected and the pages are requested to exit. This way, I in fact still push the `g_app_state` to `APP_STATE_SUBMENU`, however [input_device.c#L201](https://github.com/Master92/hdzero-goggle/blob/e82043efc21fd9fee1f71d57226a7cfee5a815a4/src/core/input_device.c#L201) instantly overwrites it.

PR #342, opened 2023-09-04
OLED burnin and static image protection

SumolX:
Performs two levels of orbiting the OSD in order to minimize screen burn-in and static image retention.

image

### DIFF
--- a/src/core/osd.c
+++ b/src/core/osd.c
@@ -686,9 +686,9 @@ static void create_osd_scr(void) {
         lv_obj_clear_flag(scr_main, LV_OBJ_FLAG_SCROLLABLE);
         lv_obj_clear_flag(scr_osd[i], LV_OBJ_FLAG_SCROLLABLE);
         if (i)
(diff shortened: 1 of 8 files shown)


SumolX:
Noticed some graphical anomalies when switching between live and menu.... setting PR to draft.

stefan74xxx:
@SumolX: Just a small question - does this orbit-function "only" work on HDZero-ui elements or although for the BetaFlight  ones (like battery voltage)? Because the last ones will although "burn in" after some time. Sorry, tried to understand the code, but not fully sure...

stefan74xxx:
Thanks for the response! Will see if i can successfully compile a firmware and give it a shot ...

stefan74xxx:
Hi - just an update: HDZero and BetaFlight OSD elements are moved by the orbit function - as expected by your explanation.

Another idea to reduce the OLED burnin problem even more "OLED burnin: Change color of OSD-elements in Google Menu" as issue #383

Issue #327, opened 2023-08-05
Enhance audible notifications -for discussion


pitts-mo:
Since I received my goggle I have wanted enhanced audible notifications and controls for audible goggle feedback. However, after PR #312 I suspect more people may have desires/suggestions for audible feedback handling/behavior . So I wanted to open this issue with some ideas for discussion.

I am currently thinking audio notifications should have:
1. priority values for all audio notifications.  Say 0-5(?) with highest number lower being lowest priory.
2. menu option/slider to select the desired priory level for audio feedback. From Critical(0) to All(5).
3. set zero priority by default to force all existing notifications to produce feedback until they are re-coded for a reduced priority level
4. add a volume control mechanism for notifications with priority >=1 
5. menu option/slider to select volume 0-10(?).

It may also be desirable to create different beep types or distinct beep types at different priorities?

pitts-mo:
Some examples of where I desire audible feedback from goggle:
-Boot complete (*&*)
-record start
-record stop
-wake from Go Sleep! (*)

maybe:
-enter sleep
-record failure?
-SD card failure?

* = PR312 incidentally adds this for V2 goggle users 
&* = only if AnalogRX Power is not needed after boot completes (v2 goggle)

evilC:
The following may benefit from audio notifications
#338 Beep when switching to FPV view is aborted
#337 Beep when long press action is triggered
In both cases, this would let the user know when the button can be released, as the desired action has taken place

PR #348, opened 2023-09-11
Add shortcut for entering sleep mode from video

### LINKED ISSUE #223: GoSleep! UX improvements

Master92:
This PR extracts sleep mode code from the UI page to its own compilation unit, allowing it to be entered from different parts of the code.

I also added the shortcut of double-clicking the right button in video mode (not DVR playback) to enter sleep mode directly. Exitting from it should always return to the previous state.
When in sleep mode, a beep is output every thirty(-ish) seconds to remind the user that the goggles is still powered on (as suggested in #223) - this feature is open for discussion and can easily be removed if unwanted.

Closes #223 

Quick-and-dirty-tested with HDZero and an analog module. Power-down and re-power-up of the latter seems to be working as expected with batch 2 goggles.

Edit: Tested this PR at a local race last weekend and everything worked out pretty nicely. The only thing left is when returning from sleep to video mode, a random image is displayed until the RX chip completed booting. It doesn't prevent anything from working - it's just a minor UX flaw that I'm looking forward to fixing yet shouldn't keep this PR from being merged.


Master92:
As mentioned by @ligenxxxx on [Discord](https://discord.com/channels/880782999006883870/967637520886161438/1151090276815867904), there is a bug in the OLED protection code, interrupting GoSleep! when it is enabled.

My latest commit 73b0823 disables OLED auto-off when the goggles are currently in GoSleep! mode.

ligenxxxx:
- _lv_inv_area: detected modifying dirty areas in render
`[Error]       _lv_inv_area: detected modifying dirty areas in render         (in lv_refr.c line)`
Once when I exited sleep mode, the HDZGOGGLE app unexpectedly exited. I'm not sure if this was related.
- wake up from video mode
Reinitializing RF takes a long time. My suggestion is to let the OLED display a progress bar, just like when entering video mode from menu mode.
- go sleep frtom outside the main menu.
I think `video mode` is more appropriate because I tried to enter sleep mode during playback without success.

rquellet:
I've tested this now and it is super useful. Nice work! Suggestions:
- Beep when entering sleep to confirm - I often want to sleep the goggle when not on my head
- Make the beep quieter
- Only do the reminder beep when in sleep mode every 3-5 minutes. Every 30sec is excessive

I'm going to use this feature this week at champs and will let you know if I find any issues

Master92:
Thank you for your testing and feedback. I'm going to try and fix/implement most of your findings. 

> - Beep when entering sleep to confirm - I often want to sleep the goggle when not on my head

I think that should be discussed first. How do you think should that beep sound? A single one? Double? Short? Long? Imho there are several things to consider and there's an issue on audible notifications #327 

> - Make the beep quieter

I'm afraid, I cannot do anything about that. Maybe only a shorter beep but I can't promise anything here.

> - Only do the reminder beep when in sleep mode every 3-5 minutes. Every 30sec is excessive

Yup, I totally agree. I'm gonna change the timeout value to something in that range.

rquellet:
I think a double beep would work. Aligns with the double press gesture.

The beep used to be shorter and quieter. A later firmware version made it louder like it is today.

Please adjust the beep frequency soon! I'm using this feature successfully at the Championship event, but people are confused why my goggle is beeping so frequently :-)

Master92:
I just searched through the code history and you're right, prior to 9.1.0 the beep would only take 500us and was changed to 50ms. I hope I will be able to make this change today. 

The frequency of notification beeps was already changed yesterday. You can download the build from [the actions page.](https://github.com/hd-zero/hdzero-goggle/actions/runs/6636376376)

Good luck and have fun at champs. And mind the tunnel 😄
I hope that maybe one day I'll make it there too but I'm still too slow for making that long of a travel 😅

Master92:
Alright, I just pushed another version containing most of the requested changes. The reminder is now more like a little knocking sound of the beeper instead of an actual beep. On top of that, a progress bar appears when returning from sleep back to video mode. You can download the binary file [from the actions page.](https://github.com/hd-zero/hdzero-goggle/actions/runs/6651252177)

I quickly tested these changes with my goggles and they seem to be working just fine. But with any beta-version: Before extensive testing, anything might happen - although the only thing that might be breaking something is the progress bar. But as I mentioned before: it *seems* to be working as expected.

PR #349, opened 2023-09-12
Notify user SD Card requires formatting

SumolX:
UI Status bar will now display "Format SD Card" when card is inserted but not mounted due to not being formatted with FAT32.


ligenxxxx:
I think there's an ambiguity, Users may understand that formatting is in progress.
What about "Format Request"?

SumolX:
Per our conversation "Unsupported" makes the most sense in regards to why the SD Card has not mounted.  When user is presented with this notification, they should format the card locally on the goggles.

PR #362, opened 2023-11-02
Fix for android devices to be able to maintain connection with goggles.

SumolX:
By setting dns and wins to 0.0.0.0 android devices will be able to maintain connectivity with goggles and not disconnect due to no internet access.


evilC:
Tested, working

PR #365, opened 2023-11-04
suppoer E1 and F1 channel

ligenxxxx:
Screenshot from 2023-11-04 13-23-55
E1:  5705MHz
F1:  5740MHz

### DIFF
--- a/src/ui/page_scannow.c
+++ b/src/ui/page_scannow.c
@@ -59,12 +59,11 @@ typedef struct {
     lv_obj_t *img1;
 } channel_t;
 
-#define SCAN_CH_NUM 12
-channel_t channel_tb[SCAN_CH_NUM];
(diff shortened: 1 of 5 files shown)


src/ui/page_scannow.c
```diff
@@ -59,11 +59,12 @@ typedef struct {
     lv_obj_t *img1;
 } channel_t;
 
-channel_t channel_tb[10];
-channel_status_t channel_status_tb[10];
+#define SCAN_CH_NUM 12
```
Master92, line 62: Couldn't you declare this constant in a more global place and use it in osd.c and dm6302.c as well?

zandorsp:
I have a sugestion for the UI on scan now menu:
highband
imd6
japan

On live view all 20 channels available on dial (L1-L8,R1-R8,E1,F1,F2,F4)
and a lowband unlock option

PR #366, opened 2023-11-06
Add cpp guards and convert to #pragma once

Master92:
This is only a cosmetic change and shouldn't make the software behave any different to as it was before. Binary diff says:

``` sh
$ diff -s HDZGOGGLE_main HDZGOGGLE_new
Files HDZGOGGLE_main and HDZGOGGLE_new are identical
```

The only reason for this PR is that you effectively can't write any cpp code right now without considering which include has to declared `extern "C"` and which doesn't.


SumolX:
I would also recommend that we also standardize the include guard to use #pragma once for all files.

Master92:
> I would also recommend that we also standardize the include guard to use #pragma once for all files.

Totally agreed! I'll update this PR in the next few hours 👍

ligenxxxx:
Please solve the conflicts.

Master92:
Force-pushed due to rebase to current main HEAD resolving the merge-conflict.

PR #367, opened 2023-11-07
Make button and roller actions configurable for video mode

### LINKED ISSUE #357: Head Tracker CENTERING function on Func button. Feature request
I am so impressed by the Head Tracker implementation, the wireless solution in particular.
Recently I had the opportunity to set it up and to use it.
Though I am really missing a convenient way to center the tracker instantly, whenever desirable. The only way available today, AFAIK is entering the HT menu and choose enter the CENTER option - and in the settings I operate this is a hassle affecting the overall HT experience.

I suggest there are some factors contributing to a great HT experience:

1. Smooth and precise sensing and output generation (check - works fine)
2. Quick and easy way of HT centering at the push of a button (not there yet)
3. Quick and easy calibration (not there yet). 
Calibration may not be necessary every session, but due to temperature variations it might be.

**Suggested solution**
Allow, via goggles configuration menu, to assign the goggles Func. button to control HT CENTER and CALIBRATE - in stead of FAN control.
On a side note - so far, I never experienced a need to adjust the fan speed, especially not during a session. While the need of HT calibration and centering is urgent.

Todays Func button default is, short press - DVR record/stop, long press - FAN speed.
I suggest fully configurable options of, SHORT PRESS, double short press  - "DOUBLE CLICK", and LONG PRESS.

Personally I would configure DOUBLE CLICK to HT CENTER and LONG PRESS to HT CALIBRATE.
SHORT PRESS would remain record on/off.

Oh, and this need of easy centering and calibration - it´s not just me. Look at the Skyzone 04X HT for instance. Dedicated HT center / calibration button on the goggles, ready to use at any moment on ground or airborne. It´s there for a reason.

Master92:
This adds a page in the settings where one can freely assign the actions that are executed when a button or the roller is operated in video mode.
I also added the ability to change the top fan speed and the OLED brightness level with the roller.

When changing a channel is requested, the left short button acts as the "Enter" button, regardless of which action is configured for it.

**Note:** This changes the behavior introduced with #348 where I double-assigned the double right click for either HT Center or Go Sleep! mode. However, as this PR is not yet in the release candidate, I think, this shouldn't do much harm. Closes #357 

The available actions are for the roller:

- Switch channel
- Change fan speed
- OLED brightness

and for the buttons:

- Toggle OSD
- Main menu
- Toggle DVR
- Center HT
- Calibrate HT
- Go Sleep!
- Toggle fan speed

Setting

Roller

Button

### DIFF
--- a/mkapp/app/setting.ini
+++ b/mkapp/app/setting.ini
@@ -45,6 +45,14 @@ enable=false
 [elrs]
 enable=false
 
+[inputs]
+roller=0
+left_click=0
+left_press=1
+right_click=2
+right_press=5
+right_double_click=3
+
 [wifi]
 enable=false
 ap_ssid=HDZero
--- a/mkapp/app/setting.ini
+++ b/mkapp/app/setting.ini
@@ -50,7 +50,7 @@ roller=0
 left_click=0
 left_press=1
 right_click=2
-right_press=5
+right_press=6
 right_double_click=3
 
 [wifi]
--- a/src/core/common.hh
+++ b/src/core/common.hh
@@ -30,6 +30,8 @@ hw_revision_t getHwRevision();
 #define RIGHT_KEY_CLICK 5
 #define RIGHT_KEY_PRESS 6
 
+#define ARRAY_SIZE(x) (sizeof(x) / sizeof(x[0]))
+
 extern atomic_int g_key;
 extern atomic_int g_init_done;
 extern pthread_mutex_t lvgl_mutex;
--- a/src/core/dvr.c
+++ b/src/core/dvr.c
@@ -87,6 +87,10 @@ void dvr_update_vi_conf(video_resolution_t fmt) {
     LOGI("update_record_vi_conf: fmt=%d", fmt);
 }
 
+void dvr_toggle() {
+    dvr_cmd(DVR_TOGGLE);
+}
+
 static void dvr_update_record_conf() {
     LOGI("CAM_MODE=%d", CAM_MODE);
     if (g_setting.record.format_ts)
--- a/src/core/dvr.h
+++ b/src/core/dvr.h
@@ -22,6 +22,7 @@ void dvr_select_audio_source(uint8_t audio_source);
 void dvr_enable_line_out(bool enable);
 void dvr_cmd(osd_dvr_cmd_t cmd);
 void dvr_update_vi_conf(video_resolution_t fmt);
+void dvr_toggle();
 
 #ifdef __cplusplus
 }
--- a/src/core/input_device.c
+++ b/src/core/input_device.c
@@ -85,12 +85,6 @@ void tune_channel(uint8_t action) {
             tune_timer = TUNER_TIMER_LEN;
             tune_state = 2;
             channel = g_setting.scan.channel;
-        } else if (action == DIAL_KEY_CLICK) {
-            g_setting.osd.is_visible = !g_setting.osd.is_visible;
-            if (g_setting.osd.is_visible)
-                channel_osd_mode = CHANNEL_SHOWTIME;
-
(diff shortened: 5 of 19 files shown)


SumolX:
Very nice!  I was working on a similar approach function pointers and all.  Excellent work!!

MrMap:
The functions of this PR is brilliant and very welcome for several reasons. 
As a Head Track user I greatly appreciate the HT Center and HT Calibrate - well done.

**I just got to think about an additional HT option for configurable button assignment: HT on/off.** 
Today I use INAV logical programming to achieve the same (which also needs one radio channel). 
I think there might be multiple situations where the HT user would benefit from easily toggling HT on/off easily on the goggles.

Master92:
> **I just got to think about an additional HT option for configurable button assignment: HT on/off.** 

Good idea. I think, some feedback in the OSD would be needed as well wouldn't it? Simply toggling would be easy but without any feedback it could be pretty hard to guess what's happening.

MrMap:
> > **I just got to think about an additional HT option for configurable button assignment: HT on/off.**
> 
> Good idea. I think, some feedback in the OSD would be needed as well wouldn't it? Simply toggling would be easy but without any feedback it could be pretty hard to guess what's happening.

Quite right! Feedback is desirable.
This also applies for HT calibration; let it beep while calibration in progress (meaning don't move goggles until calibration complete).

ligenxxxx:
This branch has conflicts that must be resolved.

Master92:
Rebased onto current `main`.

evilC:
I just flashed this today, and Go Sleep does not seem to work.
To be clear, I actually flashed [this PR](https://github.com/hd-zero/hdzero-goggle/pull/368) because I wanted both of these things - so conceivably it may have crept in as part of that.
When in the main menu, Go Sleep seems to do absolutely nothing.
If I configure LP of right button as a shortcut for Go Sleep, and then activate it while on the FPV view (But with no connection currently active - ie quad is off), then the following happens:
1. Screen off for a second or 2
2. Screen on and VERY brief loading bar
3. Screen off for about a second
4. Screen on, back to normal

Master92:
Huh, that seems strange. To be fair, I never used HDMI out, yet. Therefore I don't know what the behavior is or was supposed to be.

By saying you wanted to enter sleep mode from the main menu you mean via the lowermost menu entry, right? Because shortcuts and button actions only work in video mode.

Can you confirm that the behavior is similar with a previous release version? If that's the case, please open a new issue.
This PR didn't actually change any code related to sleep mode, therefore I can only imagine it being an already existing issue. The only time I touched sleep code was in #348 while I extracted it from one place to another.

evilC:
That was a complete red herring...
Of course, when in the menu, if you activate sleep, and you are on HDMI (Which does not disable during Go Sleep), then "Nothing happens" is how it is going to be perceived
I never used Go Sleep before, so I cannot comment on whether it changed.
TLDR - nothing to do I don't think. As soon as I realized that sleep mode does not stop feeding HDMI, all the observed behaviour makes complete sense.

pitts-mo:
Hello @Master92 
Are you able to make the input page display the current settings on fresh power up?

I have been running this from the SD card for about a week and find this a very helpful feature. The only concerning observation I have so far is that on fresh boot the button options within the input menu appear to show default settings. However, testing the functions after a power cycle does show my button preferences set during the last power session were saved correctly. Additionally, if I adjust settings my preferences the states displayed are again preserved within the input menu for the remainder of the power session.

Master92:
Ooff, I thought I already did that. So you're saying, everytime you power up, the settings page displays the defaults instead of the current settings but the button actions act as they were configured in the previous session?

I am currently unable to test that but if you can safely verify that that's the case, please open an issue and tag me so we can easier track what's going on.

Happy new year btw. 🎉😃🚀

PR #368, opened 2023-11-07
Improve main menu navigation experience

### LINKED ISSUE #363: Menu highlight should change when in sub-menu
When on the main menu, when you roll the wheel, the selected entry in the main menu is highlighted grey to indicate that the current sub-menu is selected.
When you click to enter that sub-menu, the highlight of the main menu remains grey and the sub-menu also has a highlight, which is also grey.
I feel that the main menu highlight should change in some way (Maybe turn green?) to indicate this state.

All too frequently I find myself accidentally long pressing once too many times and exiting to the FPV view, and I think this is why

Master92:
When navigating the settings, the user has no feedback on if he/she is currently navigating the main menu or a sub menu except for the highlighting of elements in the submenu.

This PR changes the behavior of the navigation bar when entering and exiting a submenu element.
I.e. when navigating the main menu, no red arrow is shown. It only appears when entering a submenu. At this moment, the highlighting of the navigation bar changes, indicating a change in the level of navigation.

This feature is open to discussion, any opinion about it is highly appreciated! When accepted, it closes #363 

#### Before

Before_main
Before_power

#### After

After_main
After_power


SumolX:
Would it be possible to modify the side menu selected entry to make it either bold or increase the font size slightly?

Master92:
> Would it be possible to modify the side menu selected entry to make it either bold or increase the font size slightly?

Yes and no.
Although I'd prefer making the selected tab bold, it is unfortunately impossible without declaring a completely new font bitmap. Slightly increasing the font size is possible, yet unfortunately looks really weird imho.

FontSize

ligenxxxx:
In the past I thought the red arrow was redundant. Now it seems to have value. It seems to mean "Look to the right (submenu)".

ligenxxxx:
There are two ways to exit the secondary menu:
- Press  `< Back` or up and left-click. After my test, there is no problem.
- Enter video mode. For example, automatically enter a certain channel after scanning. Or enter av in through source. At this time, the navigation triangle is not hidden. Please resolve this issue.

PR #369, opened 2023-11-12
Improve configuration user experience

Master92:
## Description

This PR improves the user experience when configuring the goggles. It adds kind of an `enabled` state for configuration items. That means, when an item is configured to be disabled, navigation skips this item and its text is displayed in a darker shade.
Further, when the first item in a page is disabled or only a label (for which the selection item should be set to disabled), it is skipped when entering that page.

## What pages have been changed

In detail the following pages' configuration items are affected:

- Power
	- `Battery` cell label is not selectable
	- `Cell Count` slider can only be modified if `Cell Count Mode` is `Manual`
- Fans
	- `Side Fans` slider can only be modified if `Auto Control` is `Off`
- Record Option
	- `Audio Source` can only be switched if `Record Audio` is `On`
- Auto Scan
	- `Default` selection spans two rows. They are now both highlighted as if they were one
- ELRS
	- When `Backpack` is `Off`, no action can be taken in this page except for turning it on
- Wifi Module
	- When in `AP` mode, `DHCP` and `DNS` cannot be switched
	- When in `Client` mode, `RF Channel` can never be modified and ~~`IP`-related~~ `Netmask` and `Gateway` settings can only be changed if `DHCP` is `Off`
- Head Tracker
	- When `Tracking` is `Off`, no action can be taken in this page except for tuning it on

### Edit 2023-11-30:
- Storage
	- When `develop.sh` or `HDZGOGGLE` files are detected on sd card, all options are disabled

	
## Example

The Fans page as an example:

Disabled
*Disabled `Side Fans` slider*

Enabled
*Enabled `Side Fans` slider*

### DIFF
--- a/src/ui/page_wifi.c
+++ b/src/ui/page_wifi.c
@@ -458,35 +458,56 @@ static void page_wifi_update_current_page(int which) {
         break;
     case 1:
         pp_wifi.p_arr.max = page_wifi.page_2.row_count;
-        if (page_wifi.page_1.mode.button.current == WIFI_MODE_STA) {
-            btn_group_show(&page_wifi.page_2.dhcp.button, true);
-        } else {
+        btn_group_show(&page_wifi.page_2.dhcp.button, true);
+        btn_group_enable(&page_wifi.page_2.dhcp.button, page_wifi.page_1.mode.button.current == WIFI_MODE_STA);
+        if (page_wifi.page_1.mode.button.current != WIFI_MODE_STA) {
             lv_obj_clear_flag(pp_wifi.p_arr.panel[1], FLAG_SELECTABLE);
         }
+
+        lv_obj_clear_flag(page_wifi.page_2.ip_addr.label, LV_OBJ_FLAG_HIDDEN);
+        lv_obj_clear_flag(page_wifi.page_2.ip_addr.input, LV_OBJ_FLAG_HIDDEN);
+        lv_obj_clear_flag(page_wifi.page_2.netmask.label, LV_OBJ_FLAG_HIDDEN);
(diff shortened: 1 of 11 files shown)


src/ui/page_wifi.c
```diff
@@ -455,15 +459,57 @@ static void page_wifi_update_current_page(int which) {
...
+                (page_wifi.page_1.mode.button.current == WIFI_MODE_STA &&
+                page_wifi.page_2.dhcp.button.current == 1)) {
+            lv_obj_clear_state(page_wifi.page_2.ip_addr.label, STATE_DISABLED);
+            lv_obj_clear_state(page_wifi.page_2.ip_addr.input, STATE_DISABLED);
+            lv_obj_clear_state(page_wifi.page_2.netmask.label, STATE_DISABLED);
+            lv_obj_clear_state(page_wifi.page_2.netmask.input, STATE_DISABLED);
+            lv_obj_clear_state(page_wifi.page_2.gateway.label, STATE_DISABLED);
+            lv_obj_clear_state(page_wifi.page_2.gateway.input, STATE_DISABLED);
+        } else {
+            lv_obj_add_state(page_wifi.page_2.ip_addr.label, STATE_DISABLED);
```
SumolX, line 484: IP Address can be set for DHCP mode. It allows for client to request that specific address to be reserved if DHCP server has not leased it out. Allows for DHCP mode to have a semi-static ip address.

Master92: Oh, I was unaware that this is a use-case. I just reverted this disabling part.

Master92:
> page_storage: When `HDZGOGGLE` or `develop.sh` is detected in the SD card, all storage settings are invalid.

Fixed 👍

Issue #359, opened 2023-10-27
Feature Request - When RTC mod present, place clips in date folder


evilC:
This would be useful because when viewing in the goggles, dates are not shown by the names (And is probably not really feasible given space constraints), so that if you wish to find something from a given day, it's often time consuming. The wheel is a bit laggy at best when scrolling through large amounts of clips, and this would probably help.

MrMap:
I think it might be nice with the date (and maybe time) presented with the file names, if doable.
This feature could be made user configurable.

Master92:
The file naming can be selected to represent the creation date and time since #379 has been merged. 

Placing the files in date-folders should be an easy task (that I just implemented on a private branch). Displaying the files sorted in folders in the playback page is a bit more difficult and thus time consuming. I'll try and work something out when I've got time and little by little over the next few days/weeks. I'll let you know when I've got something to try out 😉

evilC:
@Master92 are you saying that whilst they will be in folders on the SD card, that the folder structure will not be visible in the Playback menu? It will still be one big list?

Master92:
No what I'm saying is that right now (in the intermediate development state), the files get placed into folders and don't appear in the list at all. This has to be fixed so that the structure can be navigated before I'll even post a pull request 😅 Both things have to be implemented before this feature can be called usable.

Issue #123, opened 2023-01-26
Enable RTC functionality 


Wikiwix:
This issue is a container for making use of the RTC.

### Situation statement

A first step is to allow setting the datetime (and potentially the timezone) and having a secure way to determine a proper datetime is set. This is not a given as there is some tinkering involved to install an RTC battery (see [this video by @rquellet ](https://www.youtube.com/watch?v=p4qXiQrtCkU&t=443s) for details).

### Potential use cases

* DVR recordings can contain metadata with the recording datetime #116 
* DVR files could potentially contain a timestamp #42 
* The OSD menu could show the time (not sure about the value)

### Implementation suggestion

* Provide a settings entry to set time
  * This should allow setting
    * year, month and date
    * hours and minutes
    * Optionally set a time zone (otherwise the metadata will only be correct if the set datetime is given in UTC)
  * The setting should have an asterik (`*`) like the WiFi setting currently has. The footer should contain the warning
    > Time will be reset on reboot unless RTC battery is installed!

    If there is a simple way to check whether an RTC battery is installed, the asterik could be shown conditionally with a message like

    >  Time will be reset on reboot because no RTC battery is installed!
  * Currently that menu entry fits probably best in `Record Options` (already quite crowded) or `Firmware` (this could be renamed to `System` to be a little more generic)
* Features that want to use the datetime should check for the following precondition (pseudo code) to ensure a plausible datetime is configured
  ```python
   DateTime.now() >= thresholdDateTime
  ```
  `thresholdDateTime` has to fulfil the following conditions (`2023-01-01T00:00:00Z` might match them)
     * The lower bound for `thresholdDateTime` should be a date older than the day this feature ships
     * The upper bound for `thresholdDateTime` should be at least some days after the datetime the RTC starts at after loosing power
     * The lower and upper bound together ensure that metadata is only set if a plausible datetime is configured

bkleiner:
Lovely! Thank you.
i like that there is pre-planning for informing the user. feels like a good compromise to me.
Do you think you would be able to put forward a PR implementing this?

rquellet:
Here's an interesting workaround: what if we use ELRS backpack to push the RTC timestamp from radio to hdzero goggles?

bkleiner:
> Here's an interesting workaround: what if we use ELRS backpack to push the RTC timestamp from radio to hdzero goggles?

does not change much in actuality, as this is not something that we can rely on being preset.
[...]

PR #379, opened 2023-11-24
Add date-based naming scheme for DVR files and add video metadata

### LINKED ISSUE #42: Feature request: Improve the DVR filename convention
Please change the video files naming convention for something like `2023-01-24 16h46 HDZ_0042.ts` (in the VRX as well), so that the files could be ordered chronologically (alphabetically), even if they come from multiple HDZero goggle (which would then not have the same index counter).

Also, make sure the extension is `.ts`, not the uppercase `.TS`.

### LINKED ISSUE #116: Feature request: Support timestamp in DVR metadata
As discussed by @rquellet in [this YouTube video](https://www.youtube.com/watch?v=p4qXiQrtCkU&t=443s) there is a RTC in the goggles.
Having this information in the DVR files is essential for me and allows easier management of flight footage.

See #123 for details on implementation.

Master92:
This PR allows users with an active RTC to chose between the conventional naming scheme (i.e. hdz_000, hdz_001 etc.) and a date-based one (i.e. 2023-11-24T13_34_37).
When no rtc battery is present, the user can not change the naming scheme. On top of that, it automatically falls back to the conventional scheme as default.

Further, I added some metadata to the DVR files i.e. a title (currently the file name) and - for mp4 - the date it was created.
Over the course of my research and the implementation, I found out that the ts container does not support any metadata information about the media creation date, hence a note in the sources.

Closes #42
Closes #116


SumolX:
I was playing with this feature on the emulator and noticed I'm unable to change the Naming Scheme option because there is no RTC battery installed on the Emulator.  Do you think this option should be be disabled in a similiar fashion to your other PR, if no RTC battery is present then the row is not selectable and has a darker shade?

Perhaps mark the PR as a draft until the other PR has been merged, that way it follows the standard you have set forth.

Master92:
> I was playing with this feature on the emulator and noticed I'm unable to change the Naming Scheme option because there is no RTC battery installed on the Emulator. Do you think this option should be disabled in a similar fashion to your other PR, if no RTC battery is present then the row is not selectable and has a darker shade?

You're right. I already had that in mind but wasn't sure how to correctly go ahead and mark this PR to lack some work after #369. I'll convert this PR to draft in a minute 👍 
Maybe I'll play around a little and try to come up with an RTC mock for emulator builds 😉 

> Actually I could see us temporary blinking the Date/Time in live view when the recording is started.... that way we don't need to have it on permanently.... but that would potentially be a future feature.

Yeah we should keep that in mind. Perhaps you'd need the user to decide how they want the OSD to behave.

> Final comment... though its more of a personal preference.... i think we should simplify the timestamp convention to something a little more readable perhaps...
> 
> 2023-11-24T13_34_37.mp4 to 20231124_133437.mp4

That's actually a good idea imho. That way you could navigate your files easier by just typing the digits instead of having to add the dashes, too. The only thing I'd change is separating the date from time with a dash instead of an undesrscore like so: `20231124-133437.mp4`. What do you think about that?
I came up with the current scheme through [Wikiwix's comment](https://github.com/hd-zero/hdzero-goggle/issues/123#issuecomment-1466990455) in #123.

Master92:
That was a pure coincidence, they aren't on my (German) keyboard 😆

evilC:
Liking it, great!  
Any thoughts, however, on #359 ?

evilC:
It's worth noting that it appears that this PR also changed hot clip naming?  (At least I assume it was this PR, I see that `mark_video_file` changed.  
Previously, if you had no hot clips and hot clipped DVR file `hdz_028`, it would become `hot_hdz_001`, but now it becomes `hot_hdz_028`.  
I kind of like it to be honest, it helps you know where the file was in the sequence of DVR files - without the RTC mod, this was impossible to know for sure with the old system.  
It also seems that the next clip will be `hdz_029` - ie `hdz_028` would not get used again, so you would not end up with a `hot_hdz_028` and a `hdz_028`, so even if this was unintentional, I see no real need to change it back to how it was.

Master92:
That was indeed unintentional 😅

evilC:
@Master92 If we do intend to keep it as it currently is, then we should probably remove [unused code](https://github.com/hd-zero/hdzero-goggle/blob/main/src/ui/page_playback.c#L322)

Master92:
@evilC agreed. I'm currently working on that file for the date-folder feature and will try to keep removing that in mind 👍

evilC:
@Master92 I could maybe get rid of it as part of PR #389 ?

Master92:
@evilC I'd be okay with that as well 😉👍

PR #380, opened 2023-11-29
More relaxed hdmi in fps detection

ligenxxxx:


ligenxxxx:
More than 10 beta testers tested this firmware and confirmed it to have no issues.

PR #385, opened 2024-01-06
Fix initial roller function setting item

### LINKED ISSUE #384: Input menu does not display currently selected option(s) after boot (master post PR #367)
While testing master from SD card over the last week I have found #367 very helpful.

However the input menu appears to display default options after the goggle has been power cycled.

my test example:
Boot goggle With develop.sh HDZGOGGLE built from master on SD card.
-> Input menu
-> Roller
-> change option value from "Switch channel" (default) to "OLED brightness"
-> back (exit to main menu)

Initial verification(s):
-Exit to live view to verify roller does adjust OLED brightness.
-Return to the Input menu and verify OLED brightness is still displayed in the Roller field.

Power cycle goggle and re-verify:
-From live view verify roller does adjust OLED brightness.
-Return to the Input menu and verify the value displayed in the Roller field is incorrectly "Switch channel".

I believe some other Input fields and selections are affected but I have only verified the above.

Master92:
Fixes #384 

@pitts-mo Can you please verify that this PR fixes your issue? I did a quick code review and found this line that should cause it.
@ligenxxxx This PR should be included in the next release, or at least as soon as #367 is released as well.

### DIFF
--- a/src/ui/page_input.c
+++ b/src/ui/page_input.c
@@ -185,7 +185,7 @@ static lv_obj_t *page_input_create(lv_obj_t *parent, panel_arr_t *arr) {
 
     create_label_item(content, "Roller:", 1, ROLLER, 1);
     pageItems[ROLLER] = create_dropdown_item(content, rollerOptionsStr, 2, ROLLER, 320, row_dsc[ROLLER], 2, 10, LV_GRID_ALIGN_START, &lv_font_montserrat_26);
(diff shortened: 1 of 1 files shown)


pitts-mo:
I built #385 and ran from SD card to find the input page is properly displaying the set Roller option after power cycle.
Thanks,
-p

PR #389, opened 2024-01-24
Mark as Hot Clip from live view

### LINKED ISSUE #341: [Feature Request] Long press of Func button breaks recording and marks as Hot Clip
Currently, when Auto Record is on, a short press of the Func button stops recording and then recording re-starts. A long press changes fan speed, which I never use.
What I would like is for a long press of the Func button to break recording as short press currently does, but also mark the previous clip as Hot

evilC:
Fixes #341
This feature allows the user to set a shortcut via the Input menu (Called "Mark Hot Clip") that when pressed, will:
1. Stop recording if currently recording
2. Mark the last clip as a Hot Clip
3. Restart recording if we were originally recording

### DIFF
--- a/src/core/dvr.c
+++ b/src/core/dvr.c
@@ -18,6 +18,9 @@ bool dvr_is_recording = false;
 
 static pthread_mutex_t dvr_mutex;
 
+int walk_sdcard();
+void mark_video_file(int const seq);
+
 ///////////////////////////////////////////////////////////////////
 //-1=error;
 // 0=idle,1=recording,2=stopped,3=No SD card,4=recorf file path error,
@@ -180,4 +183,19 @@ void dvr_cmd(osd_dvr_cmd_t cmd) {
     }
 
     pthread_mutex_unlock(&dvr_mutex);
+}
+
+void live_mark_video_file() {
+    bool dvr_was_recording = dvr_is_recording;
+    if (dvr_is_recording) {
+        dvr_cmd(DVR_STOP);
+    }
+    int cnt = walk_sdcard();
+    LOGI("dvr_was_recording=%d, count=%d", dvr_was_recording, cnt);
+    if (cnt > 0) {
+        mark_video_file(0);
+    }
+    if (dvr_was_recording) {
+        dvr_cmd(DVR_START);
+    }
 }
--- a/src/core/dvr.h
+++ b/src/core/dvr.h
@@ -23,6 +23,7 @@ void dvr_enable_line_out(bool enable);
 void dvr_cmd(osd_dvr_cmd_t cmd);
 void dvr_update_vi_conf(video_resolution_t fmt);
 void dvr_toggle();
+void live_mark_video_file();
 
 #ifdef __cplusplus
 }
--- a/src/ui/page_input.c
+++ b/src/ui/page_input.c
@@ -36,8 +36,8 @@ typedef enum page_input_rows {
 static lv_coord_t col_dsc[] = {160, 200, 160, 160, 160, 120, LV_GRID_TEMPLATE_LAST};
 static lv_coord_t row_dsc[] = {60, 60, 60, 60, 60, 60, 60, 60, 60, LV_GRID_TEMPLATE_LAST};
 
-const char *btnOptions[] = {"Toggle OSD", "Main menu", "Toggle DVR", "Center HT", "Calibrate HT", "Go Sleep!", "Toggle fan speed"};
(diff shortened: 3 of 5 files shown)


evilC:
Please be aware that I am a total noob at C, so this is quite possibly shit code.  
I had to make `walk_sdcard()` and `mark_video_file()` in `ui/page_playback.c` not static, as I could not work out how to reference it from my code in `dvr.c` otherwise - is there a better way of doing this? Did they need to be static in the first place?

Master92:
Good job figuring that out as your first pull request, even though you claim to be "a total noob at C".

Personally, I'd implement it differently as those two methods from `page_playback` are actually meant to be only used in connection with its media database. `walk_sdcard()` could potentially become really slow over time and should really not be used during live video imho. It even copies all `.jpg` files to `/tmp` as a side effect.
Although it should work for now, it will definitely break as soon as date-folders are introduced (I'm currently working on that).

If you ask me, you should better query the recorder app for the current file name and continue renaming this file (and its thumbnail) with that information. I know that this is a bit more complicated but that's the only safe way I can think of getting the actual current dvr filename (at least for now).

evilC:
Yeah, I felt kind of dirty doing it the way I did, I suspected as much.  
TBH, thinking about it, I was approaching it from the use-case that I would use it in - I would always be landed when triggering it, so had not really thought of live view being "active" per se (Yes, the feed would be live for me, but I would not be airborne, so no real risk if it chugged the system a bit).  

WRT querying the recorder app, iI had a look and it was pretty incomprehensible to me what is going on (Or maybe I am looking in the wrong place?). Furthermore, doing it like that would only work if we are currently recording, and not if we had stopped recording. 
One idea I had was that maybe the recorder should write some variable somewhere which stores the name of the last recorded clip. This of course would mean that the shortcut would not work if you booted up and then hit it, but I don't really see a need to support that - certainly for me, the two use-cases are (1) landed, quad still powered up, and want to hot clip, and (2) landed, unplugged, and want to hot clip.

Another idea I had was to keep basically the current implementation, but avoid calling `walk_sdcard`. I have implemented this for now as it was pretty simple to do.  
I have not yet tested, however, the scenario of what would happen if there were no files on the SD card.

evilC:
> If you ask me, you should better query the recorder app for the current file name

I had a look at this and I could not see how to get this information :(  
I see [here](https://github.com/hd-zero/hdzero-goggle/blob/main/src/record/main.c#L457) that `sFile` probably holds the filename, but I tried logging it out using `LOGI` and don't see it appear in the logs. I notice that all the other log statements in the file are `LOGD` or `LOGE`, so I wonder if there is any other log file that I am missing? `recording...` [should be logged out at debug level somewhere](https://github.com/hd-zero/hdzero-goggle/blob/main/src/record/main.c#L270), but I don't see it. Even then, I am not sure how I would export it - `main.c` does not seem to have a `main.h`, so I'm a little lost here :(

Master92:
Yep, recorder is a separate application that runs on its own. Therefore, it logs somewhere else, yet I also don't know where 😅
Hm - second thought is maybe add a command to the recorder that does what you intend to do and issue that from the HDZGOGGLE application just like start and stop commands. Even though it might duplicate some code 😕

SumolX:
@evilC perhaps mark this PR as a draft until it is ready in order to prevent an accidental merge into main before you feel it's ready.

evilC:
> @evilC perhaps mark this PR as a draft until it is ready in order to prevent an accidental merge into main before you feel it's ready.

Agreed. I have had to shelve this for the time being as it seems that it is a little more involved than originally thought

qvasic:
A couple of days ago I had an idea for similar feature so it is nice to see that somebody is already doing it. :)
I'll take a look at your code and test it.

evilC:
> A couple of days ago I had an idea for similar feature so it is nice to see that somebody is already doing it. :) I'll take a look at your code and test it.

The existing code will not work. My initial code did, but called `walk_sdcard`, which is probably a bad idea.
I need to re-visit it at some point, but have been a bit busy lately

suzrik:
Are you going to finish it in the near future? Maybe I can help you with this?

evilC:
@suzrik I have a bunch going on at the moment which means I don't really have much time to devote to it. If you wanna pick it up and run with it, feel free :)

suzrik:
@evilC Thank you, I'll take it on :)

qvasic:
> @evilC Thank you, I'll take it on :)

@suzrik, there is already similar functionality merged in: https://github.com/hd-zero/hdzero-goggle/pull/394

suzrik:
@qvasic your idea is better. Thank you!

Issue #391, opened 2024-01-28
Emulator crashes immediately on Linux (debian bookworm)


lukolszewski:
The emulator crashes instantly on startup, but after many tries it did eventually start once (out of maybe two dozen tries). It crashes in /src/core/input_device.c:485, where SDL_WaitEvent(&event) is called in a new thread context at the same time another SDL activity is happening in the main thread. 

The call happens in a function called thread_input_device(void *ptr) which is called from input_device_init like this: pthread_create(&input_device_pid, NULL, , NULL);

Using multithreading with SDL like this will result in crashes(only one thread can use SDL at a time). Perhaps a SDL_mutex can be used to coordinate use of SDL between the threads in the emulator. 

Here is the crash output as run with gdb:
Starting program: /home/xxxxx/hdzero-goggle/build_emu/HDZGOGGLE
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff6bfe6c0 (LWP 1579019)]
[INFO][rtc.c:rtc_init:186] rtc_init has NOT detected a battery
[INFO][rtc.c:rtc_init:190] rtc_init updating both clocks via settings
[New Thread 0x7ffff63fd6c0 (LWP 1579020)]
[New Thread 0x7ffff5bfc6c0 (LWP 1579021)]
[INFO][mcp3021.c:aduc_read_voltage:60] Voltage ADC is mcp
[ERROR][mcp3021.c:mcp_read_voltage:31] mcp_read_voltage: open error.
[New Thread 0x7ffff53fb6c0 (LWP 1579022)]
[INFO][common.c:bmi2_interface_init:95] I2C Interface
[ERROR][common.c:bmi2_error_codes_print_result:165] Error [-3] : Device not found error. It occurs when the device chip id is incorrectly read
[INFO][accel_gyro.c:init_bmi270:74] init_bmi270 done.
[INFO][accel_gyro.c:enable_bmi270:88] [Pass] BMI270 enabled.
[INFO][fans.c:fans_top_setspeed:24] fans_top_setspeed: 4
[New Thread 0x7ffff7fc17c0 (LWP 1579023)]
[New Thread 0x7ffff4bfa6c0 (LWP 1579024)]

Thread 5 "HDZGOGGLE" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff53fb6c0 (LWP 1579022)]
[...]

PR #392, opened 2024-01-28
Fixing emulator bug #391

lukolszewski:
The emulator would crash for me. This is either because SDL was not initialised before SDL_WaitEvent was called, or more likely because anothe thread was trying to use SDL at the same time. This fixes both by initialising SDL earlier and by setting up a mutex that locks access to SDL when the input thread works. 

**This affects the emulator only, it has no impact on running on the goggles**

### DIFF
--- a/src/core/input_device.c
+++ b/src/core/input_device.c
@@ -14,6 +14,7 @@
 #include <minIni.h>
 
 #ifdef EMULATOR_BUILD
+#include "SDLaccess.h"
 #include <SDL2/SDL.h>
 #endif
 
@@ -482,7 +483,9 @@ static void *thread_input_device(void *ptr) {
 
     while (true) {
         SDL_Event event;
-        while (SDL_WaitEvent(&event)) {
+        SDL_LockMutex(global_sdl_mutex);
+        while (SDL_PollEvent(&event)) {
+            SDL_UnlockMutex(global_sdl_mutex);
             switch (event.type) {
             case SDL_QUIT:
                 exit(0);
@@ -539,7 +542,10 @@ static void *thread_input_device(void *ptr) {
                 }
                 break;
             }
+            SDL_LockMutex(global_sdl_mutex);
         }
+        SDL_UnlockMutex(global_sdl_mutex);
+        usleep(50000); // Sorry, this will break windows, but it's not like it is working now anyway :-(
     }
 #endif
 }
@@ -561,5 +567,8 @@ void input_device_init() {
     }
     app_state_push(APP_STATE_MAINMENU);
 #endif
+    if (SDL_Init(SDL_INIT_VIDEO) != 0) {
+        printf("Error initializing SDL: %s\n", SDL_GetError());
+    }
     pthread_create(&input_device_pid, NULL, thread_input_device, NULL);
 }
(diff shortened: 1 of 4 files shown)


src/core/input_device.c
```diff
@@ -14,6 +14,7 @@
 #include <minIni.h>
 
 #ifdef EMULATOR_BUILD
+#include "SDLaccess.h"
 #include <SDL2/SDL.h>
```
SumolX, line 18: Don't need to include this header since its already included via sdlaccess.h

src/core/main.c
```diff
@@ -11,6 +11,12 @@
 #include <lvgl/lvgl.h>
 #include <minIni.h>
 
+#ifdef EMULATOR_BUILD
+#include "SDLaccess.h"
+#include <SDL2/SDL.h>
```
SumolX, line 16: Same issue here.  No need to include SDL.h here either.

src/core/main.c
```diff
@@ -141,6 +147,13 @@ void lvgl_init() {
 int main(int argc, char *argv[]) {
     pthread_mutex_init(&lvgl_mutex, NULL);
 
+#ifdef EMULATOR_BUILD
+    global_sdl_mutex = SDL_CreateMutex();
+    if (global_sdl_mutex == NULL) {
+        // Handle error: SDL_CreateMutex failed
```
SumolX, line 153: Please add a log entry.  Good to know if it ever fails.

src/ui/ui_porting.c
```diff
@@ -10,6 +10,7 @@
 #include "lvgl/lvgl.h"
 
 #ifdef EMULATOR_BUILD
+#include "SDLaccess.h"
 #include <SDL2/SDL.h>
```
SumolX, line 14: Please remove SDL.h here as well.

src/ui/ui_porting.c
```diff
@@ -127,8 +130,11 @@ int lvgl_init_porting() {
 
     lv_disp_drv_init(&disp_drv);
 #else
-    SDL_InitSubSystem(SDL_INIT_VIDEO);
+    if (SDL_WasInit(SDL_INIT_VIDEO) == 0) {
+        SDL_InitSubSystem(SDL_INIT_VIDEO);
+    }
```
SumolX, line 135: maybe add else { log it (not that it should ever happen) }

src/core/input_device.c
lukolszewski: done

src/core/main.c
lukolszewski: done

src/ui/ui_porting.c
lukolszewski: done

src/ui/ui_porting.c
lukolszewski: Added. It may happen if the input thread managed to initialise first.

lukolszewski:
Thanks for the review, all changes were added..

src/core/main.c
lukolszewski: done

ligenxxxx:
@lukolszewski Build firmware failed. Please fix this.

SumolX:
@lukolszewski Is the PR abandoned?  If not please do update the branch and we can get it merged into main.  Thank You.

eliasrenman:
I went ahead and fixed this over at #392 so please have a look there 😀

PR #395, opened 2024-02-20
Make a beep once goggles are up and running.

qvasic:
HDZero Goggle take around 20 seconds to start up.
This PR makes goggles beep once they started up, so user doesn't have to check visually.

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -193,6 +193,7 @@ int main(int argc, char *argv[]) {
 
     // 10. Execute main loop
     g_init_done = 1;
+    beep_dur(BEEP_SHORT);
     for (;;) {
         pthread_mutex_lock(&lvgl_mutex);
         main_menu_update();


IDV7:
Suggestion: make it an option in the gui?

ligenxxxx:
> Suggestion: make it an option in the gui?
Yes, there may be people who don't like beep.

qvasic:
> > Suggestion: make it an option in the gui?
> 
> Yes, there may be people who don't like beep.

I'll do it once I have some free time.

nerdCopter:
much desired option.

sanderpuh:
There's already a beep in my case when my goggles boot up to HDZero. The expansion module is set to automatically switch off if not used, and the goggles beep as soon as the power to the expansion module is toggled.

Could this feature be implemented by having the goggles do a beep when the source is changed? This means it will beep on startup, and when the source is changed, to indicate the switch was succesful.

nerdCopter:
addition of `.vscode/launch.json` should not be in this PR.

qvasic:
> addition of `.vscode/launch.json` should not be in this PR.

@nerdCopter why not? I tested it with WSL, it should work in native Linux as well. Very useful if you need to debug something.

pfeerick:
>why not? 

Because PRs are better when they focus on just one thing/area.  So while, useful, that change does not relate to the rest of the PR and its overall area of focus. It would probably be better to have that in a in a separate PR, adding just that functionality. Smaller/single focus PRs are easier and quicker to review, and thus more likely to get merged sooner. i.e. that might have been merged a couple of months ago now if it wasn't part of this PR ;)

PR #403, opened 2024-04-07
Add option to disable random MAC address in client mode

yenon:
To allow for easier access via WIFI, the changing mac address of the goggle will result in different IP's every boot, thanks to the dhcp on the network not recognizing the device by it's MAC.
This PR introduces a (semi) permanent MAC that can be used as a fixed address, if so desired.
The first address of the device is generated randomly from system time and saved to the settings_ini.
Changing of the address is also supported by placing a file named macaddr.txt in the SD-card root.
Validation is in place, and only valid mac's will be accepted.

Please validate this a bit more cautiously, my C skills are rusty.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/.cache
 /build
 /build_emu
 linux-4.9
--- a/mkapp/app/script/wlan_stop.sh
+++ b/mkapp/app/script/wlan_stop.sh
@@ -2,6 +2,7 @@
 /mnt/app/app/record/gogglecmd -live quit
 sleep 1
 killall rtspLive
+killall wpa_supplicant
 killall hostapd
 killall udhcpd
 killall dropbear
--- a/src/core/defines.h
+++ b/src/core/defines.h
@@ -1,8 +1,8 @@
 #pragma once
 
-#define DIAL_SENSITIVITY 1    // number of clicks before dial event is triggered
+#define DIAL_SENSITIVITY           1    // number of clicks before dial event is triggered
 #define DIAL_SENSITIVTY_TIMEOUT_MS 1000 // ms
-#define CHANNEL_SHOWTIME 30 // must <= 127
+#define CHANNEL_SHOWTIME           30   // must <= 127
 
 #define GPIO_BEEP 131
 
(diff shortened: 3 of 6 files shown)


SumolX:
Nice job!  I would recommend not keeping both clientid and the option for random MAC address.  Remove the clientid setting attribute and associated references, use the persistent Mac attribute instead.  Simplify the code to always used the generate MAC address instead.  I don't see a benefit for having the user switch between random and static MAC address and having extra files on the sd card since if the card gets formatted this MAC address is lost.  Best to keep it all embedded within settings framework and the clientid becomes deprecated in favor for mac attribute.

yenon:
Before removing features, I'd like to know nobody actually uses them.
Secondly, what does clientid actually do? I cannot find a manpage for udhcpc that has the x flag, no idea what it does.
Removing random MAC completely might be bad for privacy in open networks, thanks to fingerprinting. (We are talking about a video goggle though.)

I see the point of not needing the MAC to be changed via SD card. Mostly just me wanting to set the default to what it used to be on my first try. Maybe in case of a settings file update?
I foresee no real changes to the way MAC addresses work in the wild, so the code would essentially remain untouched for a long time, so why not have it? How it works is actually described in the note at the bottom of the settings page.

yenon:
Keeping the MAC static seems sane for me, especially with our usecase. The clientid can be the static mac address, just as a fallback in case of weird behavioured dhcp's. I'll drop the menu entry for changing it, that will already simplify a lot. I'd like to keep the mac changing by file, since we do not adhere to any reasonable standard for it's assignment and duplicates are unlikely, but not impossible. Sound good?

yenon:
Should be alright now.

nerdCopter:
April 26 comment make 100% sense to me as someone in the IT industry. Maybe rebase on `main` and mark ready for review.

yenon:
It would make sense, if the used wifi chipset actually supported it properly. The current driver seems to offer better performance, but lacks any means of changing the mac address, if I remember correctly. Feel free to pull and build this, but it will not work.

SumolX:
I would close this PR if a new WiFi driver is not included since there is a driver limitation that prevents this PR from being successfully integrated.

jwhammer:
A fixed MAC address in Wi-Fi client mode is required in order to assign a static IP address via DHCP on the router.  
With random MACs this is not possible, which makes DHCP reservation unusable for the goggles.  

Also, in the current firmware (9.5.1 Rev20250715) static IP mode without DHCP does not work,  
so DHCP reservation is the only practical option – but this requires a stable MAC address.  

Please consider merging or reviving this PR.

PR #418, opened 2024-06-04
Follow DCF for DVR storage

sanderpuh:
- Store DVR files in folder /DCIM/100HDZRO/
- Extend index length to 4 (hdz_xxx to hdz_xxxx)

[Design rule for Camera File system (DCF)](https://en.wikipedia.org/wiki/Design_rule_for_Camera_File_system#Directory_and_file_structure)

### DIFF
--- a/src/record/record_definitions.h
+++ b/src/record/record_definitions.h
@@ -89,7 +89,7 @@ extern "C" {
 #define REC_packTypesNUM    2
 
 #define REC_filePathGet(BUFF, PATH, PREFIX, INDEX, FILEFMT) \
-    sprintf((BUFF), "%s%s%03d.%s", (PATH), (PREFIX), (INDEX), (FILEFMT));
+    sprintf((BUFF), "%s%s%04d.%s", (PATH), (PREFIX), (INDEX), (FILEFMT));
 
 #define ZeroMemory(p, size) memset(p, 0, size)
 
--- a/src/record/record_definitions.h
+++ b/src/record/record_definitions.h
@@ -71,7 +71,7 @@ extern "C" {
 #define REC_packSIZE        1024                //MB
 #define REC_minSIZE         50                  //MB
 #define REC_maxSIZE         (2 * 1024)          //MB
-#define REC_packPATH        "/movies/" //REC_diskPATH "/movies/"
+#define REC_packPATH        "/DCIM/100HDZRO/" //REC_diskPATH "/DCIM/100HDZRO/"
 #define REC_packPREFIX      "hdz_"
 #define REC_hotPREFIX       "hot_"
 #define REC_packHotPREFIX   REC_hotPREFIX REC_packPREFIX
@@ -82,7 +82,7 @@ extern "C" {
 #define REC_packPGN         "png"
 #define REC_packTYPE        REC_packTS
 #define REC_packSnapTYPE    REC_packJPG
-#define REC_packIndexLEN    3
+#define REC_packIndexLEN    4
 #define REC_packTYPES       {REC_packMP4,REC_packTS}
 #define DOT                 "."
 #define REC_packEXTS        {DOT REC_packMP4, DOT REC_packTS}


ligenxxxx:
`#define REC_filePathGet(BUFF, PATH, PREFIX, INDEX, FILEFMT)`
Does `INDEX` need to be changed to `%04d`?

ligenxxxx:
When the file index is 9999, what is the next index? I don't seem to find the code to clear the index to 0, although it is not the content of this PR.

sanderpuh:
> `#define REC_filePathGet(BUFF, PATH, PREFIX, INDEX, FILEFMT)` Does `INDEX` need to be changed to `%04d`?

Good call, overlooked that one. Yes, `INDEX` needs to be changed to `%04d`.

Regarding the next index after 9999 - I don't think we'll ever get to that point, so it is acceptable for now. DVR files have a large size, SD card will be filled long before we hit 9999 files. Using DCF in general allows us to review DVR files on iOS. 

In a future release we can further enhance the DCF format. Each DCF directory may contain up to 9999 DCF objects, numbered from "0001" to "9999". This number is also known as a "file number". First DCF directory starts with 100, in my code it's hard coded as "100HDZRO". For example, the next file after hdz_9999 should create a new folder "101HDZRO", hdz_0000 again. So it would also require an index for the directory.

ligenxxxx:
@sanderpuh 
image
You should also change this, otherwise the folder will not be created.

ligenxxxx:
If users use time to name video files, can they satisfy DCF?

sanderpuh:
> @sanderpuh image You should also change this, otherwise the folder will not be created.

You mean this should change in case we're generating multiple subdirectories, correct? As is, the directory `/DCIM/100HDZRO/` is generated and recorded files are stored, starting with `hdz_0000`.

If users use date / time to name video files, ánd satisfy DCF, you have to stay within the rules of DCF. You could use the subdirectory and file numbering to include the date in naming. 

**Subdirectories**
The file system contains a DCIM directory, which can contain multiple subdirectories with names such as "123ABCDE" that consist of a unique directory number (in the range 100…999) and five alphanumeric characters (or any valid filename characters), which may be freely chosen.

**Files**
The directories contain files with names such as "ABCD1234.JPG" that consist of four alphanumeric characters, followed by a number in the range 0001...9999.


For example:
- DCIM
- - xxxYMMDD
- - -  HHMMzzzz
`/DCIM/xxxYMMDD/HHMMzzzz.mp4`

xxx: unique directory number **100**...999
Y: year (202**4**-06-06) 
MM: month (2024-**06**-06)
DD: day (2024-06-**06**)
HH: hour (**08**:55)
MM: minute (08:**55**)
zzzz: unique file number **0001**...9999

The final file in the above example would be `/DCIM/10040606/08550001.mp4`


NOTE: I have not tested the above directory example as DCF structure yet! This requires further testing, perhaps there's better naming structures as well.

ligenxxxx:
Currently this PR will cause the recording to not start because the DCIM/100HDZRO folder is not created correctly.
`mkdir("/DCIM/100HDZRO", 0777)` will fail because the DCIM folder does not exist.
I recommend using `system("mkdir -p /DCIM/100HDZRO")`, Of course you need to introduce variables.

The option to name the DVR by date is in the menu, please test it.

ligenxxxx:
DCF is satisfied when `Naming Scheme` is set to` Digits`.
`/DCIM/100HDZRO/hdz_001.ts`
DCF is not satisfied when `Naming Scheme` is set to `Date`.
`/DCIM/100GDZRO/20240606-175934.ts`
Do you think this is ok?

sanderpuh:
@ligenxxxx please see #422 - Subdirectory is now created correctly, and permissions set to 777.

Naming Scheme options are best of both worlds, people can then choose between DCF format or Date format. The option to change the Naming Scheme is greyed out in my build, because I don't have the RTC battery installed in the goggles, so I could not test that option (yet).

PR #422, opened 2024-06-06
Subdirectory support for disk_checkPath()

sanderpuh:


ligenxxxx:
I tested it with no problem.

PR #425, opened 2024-06-17
Decreased overall bootup time.

SumolX:
Implemented a priority list which will defer time consuming processing until the menu system is up and running.

Tested with my 256 GByte SD Card which has alot of files we have gone from 30 seconds down to 24 seconds.

With an empty 32GByte SD Card we are now averaging around 18-19 seconds.

Without SD Card we are 16 seconds.


nerdCopter:
before this commit, occasionally my goggles would seem stuck on boot screen forcing me to reboot. since running `84fdd30` not only boot is quicker, but seems i have not gotten any stalls. thank you.

PR #426, opened 2024-06-21
Clear DVR folder on storage page

ligenxxxx:
Screenshot from 2024-06-21 12-16-49


sanderpuh:
Maybe [Remove DVR folder] isnt the best phrase to use, because the folder will be created again. Perhaps it is better to use the phrase [Clear DVR folder] or [Delete all DVR content] or something similar?

PR #423, opened 2024-06-07
Quickswap sources using the Input menu options

sanderpuh:
@Master92 did quite a lot of coding already, I took inspiration from his work which allows cycling through the source options.
On top of that I've added the option to choose to cycle only between HDZero and the expansion module, which is most useful on races where people use both systems, and for people flying lots of whoops with both systems.

Added Toggle Mode
I've added `Toggle Mode` to the `Auto Scan` screen, because this seemed the most logical to me. Open to suggestions!
The `Cycle` setting does cycle through all source options, the `Race` setting switches between the expansion module and HDZero. In case `Race` is selected, and anything other than the expansion module or HDZero is active, it will switch to HDZero when the button is pressed to switch sources.

Added Toggle source
The dropdown menu on the `Input` screen now has the option `Toggle source` added to the list, in this case I've set it on the `Right long:` button (not as default).

The code also takes into account that the expansion module can be switched off automatically, which is taken care of in the quickswap too.

Issue #208, opened 2023-04-06
Goggle DVR scaling between 720p/1080p


burnitdn:
ISSUE:
DVR scales/crops the output of the screen when recording after switching resolution modes between 720p and 1080p
![1080-720 switch](https://user-images.githubusercontent.com/125995456/230474681-1f0145a7-8fd8-459f-8519-3f89304c64b7.png)
![720-1080 switch](https://user-images.githubusercontent.com/125995456/230474683-bc299043-e031-47f7-9568-b4a881086e81.png)

TO REPRODUCE:
Goggles and vtx using rev04042023 firmware
Set goggles to record manually with .ts format

Power goggles/aircraft, enter camera menu, switch aspect ratio from 16:9full to 1080p30 (or vise versa), save and exit menu, manually start recording on goggles, manually stop recording. Resulting DVR file will be scaled/cropped up or down depending on which way the aspect ratio was switched.

ligenxxxx:
solved.

PR #433, opened 2024-08-05
Add quickswap input options to inputs menu

sanderpuh:
Simplified version of the quickswap input PR, this PR replaces the old #423. 

Added 2 options to the dropdown menu on the inputs page: `Toggle source` and `Cycle source`
Added the remark 'Toggle source' will switch between HDZero and Expansion module
Everything can be set on the inputs page, without the need of defining which option you want to use on a separate page.

See `Right short:` and `Right long:` in the below example:
Screenshot 2024-08-05 174353

### DIFF
--- a/src/ui/page_input.c
+++ b/src/ui/page_input.c
@@ -11,6 +11,7 @@
 #include "core/osd.h"
 
 #include "ui/page_fans.h"
+#include "ui/page_source.h"
 #include "ui/ui_image_setting.h"
 
 /**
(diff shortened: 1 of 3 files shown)


nerdCopter:
i see it merged, but will still post.
- briefly tested "Toggle Source" -- worked as intended.
- regarding "white screen" for no input -- how does this affect the OLED screen health if left active for some time? is burn-in an issue? maybe this could be a black screen with moving text "No Input". Thoughts @ligenxxxx ?

My user experience:
- First bootup to set a button to "Toggle Source".
- Second bootup to test the button: worked as intended. no signal testing.
- Third bootup to fly: booted analog quad after booting goggle. Had a hard time locking onto the signal, including rolling signal, and a couple instances of white screen (white screens did not DVR record). I wiggled analog module but it was firm. Rebooted quad, finally after a few more attempts it locked in and i flew.
- Fourth boot: Locked onto analog instantly and i flew.

DVR of my experience, red screens indicates DVR file change. 1 minute of switching/etc before geting a good lock on analog. Also note end of video recording at 1080 on green as per Issue #208 (@4m17s)
- https://youtu.be/JrYVjZYzLt4

sanderpuh:
Thank you for your feedback @nerdCopter - I don't recognize the experience. I'll have a look though, perhaps I can find something off.

Issue #432, opened 2024-08-05
Strange DVR on latest build if VTX is on before goggles 


alfonsogordon:
i think i've narrowed this down to only when the VTX is powered up before the goggles (or before the goggles have fully booted). If i have the VTX powered on first the DVR files come out where the image is in the top left of the frame and the rest is bright green. 
![image](https://github.com/user-attachments/assets/90b10438-15e9-418e-86ae-1c1c5689087b)

alfonsogordon:
> #434 should fix this I think, unless this isn't caused by #433

Didnt seem to fix it for me

sanderpuh:
You're right, the issue you raised is from before #433 so the issue is in something else. I have the same issue, if the goggles are booted and you start recording without opening the menu first, the recording is not properly scaled

ligenxxxx:
Does this happen only when wifi is turned on?

ligenxxxx:
![image](https://github.com/user-attachments/assets/94e7b6ad-0ad8-486f-8eec-ab0bb5f8e4d4)
This is a questionable place. It may be a problem of execution order.

sanderpuh:
I have never used the wifi option on my goggles, so if the default setting for the wifi interface is `OFF`, then it also happens when wifi is turned off. 

I have been able to reproduce the behavior, my goggles are set to boot to 'last' source. If I start the recording, it does not scale the DVR. After going to the menu, and back to the source, the DVR is scaled. So perhaps if 'last' option is used for source, it skips the DVR scaling?

ligenxxxx:
@SumolX Maybe it is related to https://github.com/hd-zero/hdzero-goggle/pull/425 `page_wifi_update_settings()` is postponed but may be run in video mode.

SumolX:
@alfonsogordon and @sanderpuh could you verify with the WiFi setting enabled option set to off if you can still reproduce this issue?

chaotixtech:
I have been experiencing this issue for quite a while. Just turned wifi to OFF and rebooted the goggles, first 2 test videos are recording correctly now (eco vtx) Goggles on last commit to master and vtx 1.6.0

SumolX:
> I have been experiencing this issue for quite a while. Just turned wifi to OFF and rebooted the goggles, first 2 test videos are recording correctly now (eco vtx) Goggles on last commit to master and vtx 1.6.0

Are you running latest 9.4 firmware?  If not could you post the version reported in firmware and include the git commit id as well.

[...]

PR #438, opened 2024-08-27
Prevent recording race-condition when wifi is enabled

SumolX:
Resolves issue #432 where recording is initiated before wifi configuration has completed.  This is breakage introduced due to boot up optimization.

### DIFF
--- a/src/ui/page_wifi.c
+++ b/src/ui/page_wifi.c
@@ -261,6 +261,9 @@ static void page_wifi_mask_password(lv_obj_t *obj, int size) {
 
 /**
  * Update settings and apply them.
+ *
+ *  Note: This function will be invoked asynchronously post bootup and may
+ *        require additional APP_STATE checks to ensure integrity of execution.
  */
 static void page_wifi_update_settings() {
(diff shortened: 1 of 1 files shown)


chaotixtech:
Tested with an eco vtx, 3 flights, all correctly recorded, wifi enabled.

sanderpuh:
Flashed and tested it, goggles are set to Auto Scan -> Default: Last

When my goggles are booted up using analog, start recording, the problem still persists with the DVR scaling. After going through the menu and back to recording analog, it is properly scaled. So for my scenario it is not yet solved.

SumolX:
> Flashed and tested it, goggles are set to Auto Scan -> Default: Last
> 
> When my goggles are booted up using analog, start recording, the problem still persists with the DVR scaling. After going through the menu and back to recording analog, it is properly scaled. So for my scenario it is not yet solved.
> 
> For what it's worth, after I switched WiFi to OFF, the issue does no longer show up. Switching it back to ON, and the issue is back.

I am still working on it and will supply a proper fix.  Root cause has been identified in regards to the race condition.

nerdCopter:
fixed AFAICT

PR #439, opened 2024-09-02
Add Beta Localization Support for HDZERO Goggles Firmware

Sat1l:
## This PR introduces beta localisation features to HDZERO goggles firmware.

Screenshot 2024-09-02 at 21 25 16

## Some added things include:

- Automatically installing the [lv_i18n](https://github.com/lvgl/lv_i18n) tool within the devcontainer to compile `.yml` localization files.
- A more symbol-rich Montserrat font to support most European and some Asian languages. **Note** that this change has introduced some layout issues (see below).
- A _currently non-dropping-down_ dropdown in the Firmware tab to select languages.
- Russian language support, provided by a native speaker.

## What works:
- New languages can be added to the firmware by including a `.yml` file, compiling the languages, and rebuilding the firmware.
- Implemented a language switcher within the UI.
- Successfully loading the selected language.
- Language is stored as a setting, so after the first reboot to set the language you can just leave it.

After compiling the code with new languages, changing the firmware language is as easy as adjusting any other setting and rebooting

## Current issues that make this PR a beta:
- Only Russian has been tested, so bilingual testers are needed to verify other languages.
- Exiting clock page raises a SEGFAULT (this one is beyond my understanding).
- The battery icon in the status bar does not change color as expected.
- There is no font loader implemented, limiting the ability to add languages with complex character sets (e.g., Asian languages). _A potential workaround is to use font backups._
- Some UI labels/options data is directly tied to their displayed text, making localization more complex and requiring additional finicking.
- No guide on adding a new language, though it's almost the same as in lv_i18n readme, it would be really helpful to have our own guide and language PR guidelines

Screenshot 2024-09-02 at 21 25 26


ligenxxxx:
I like the idea of ​​putting the language options in the firmware menu instead of creating a new menu.

Sat1l:
Was able to find a person with actual hardware - firmware works and the sim issues are the same on there.

Sat1l:
I was able to identify the issue with clock page raising a SEGFAULT. 
If the "< Back" button has more then 6 symbols closing the Clock page would cause a crash, and russian translation is "< Назад" which is 7 symbols. Still how to solve this issue and hoping someone with better lvgl knowledge could help.

ligenxxxx:
Asian languages ​​are really hard. I am from China and I plan to add only Chinese fonts. This means that every time I add a string, I need to update the fonts.

Sat1l:
> Asian languages ​​are really hard. I am from China and I plan to add only Chinese fonts. This means that every time I add a string, I need to update the fonts.

Not exactly. One font in the four (4 different font sizes are used in the ui) sizes is sufficient. I haven’t experimented yet with the font loading, only used backup fonts. After adding a new string you just have to type one lv_i18n command to compile the .yml, and re-build the firmware.
Other then Russian I tested it with some German and it worked fine

Sat1l:
Screenshot 2024-11-05 at 20 19 00

I did some Deepl translation and it kinda works. I as of now I solely rely on font-backups, but its not a rigid approach in any way shape or form. 

Issues to target are:
- lagre filesizes of the cjkui block - fonts weight 200MB and some of them are in millions of rows
- creating a font-loader (so that we don't have to hardcode the waterfall of backups)

As for the first one, if it's fine with you i can add an additional command to parse all of the used symbols in translations and generate the fonts only with the needed values.

So adding a new string would look like this: create a key - extract the key - write the key pair in english - write a value in any desired language - run a font-compile command.

This way we won't have to store hundreds of megs of .c fonts in a repo, the output fonts will be quire small, and we will have to only store one font file to generate all the sizes. This could also be done during ./setup.sh.

As for the font loader, if we do the compilation approach we won't require the loader since all of the symbols would be in one font. My goggles are being shipped, so when they arrive I'll test if including large font influences load time significantly.

@ligenxxxx I'm pending your opinion!

ligenxxxx:
I plan to create PRs myself to implement multiple languages.

ligenxxxx:
https://github.com/hd-zero/hdzero-goggle/pull/452

Sat1l:
so should I continue working on this pr or from now on internalization is your job?

ligenxxxx:
> so should I continue working on this pr or from now on internalization is your job?

You can wait for a while, let me try it first.

Sat1l:
did this while finicking with branch names, if help is needed I'll create a new one

PR #457, opened 2024-11-13
Simplify post bootup action handling

Master92:
I got confused about how the post bootup actions were handled so I decided to change the approach of how the software deals with them.

First off, I could reduce the necessary array size of the bootup actions by using the stdlib `qsort` method instead of a bubble sort approach.
Secondly, instead of querying all possibly existing post bootup actions when updating the menu, I start the first action and then call the handling method again where the next action is already selected as soon as the action completes. This even works with multithreading because every action is only dispatched from the main ui thread.


src/ui/page_storage.c
```diff
@@ -584,6 +584,10 @@ static void *page_storage_repair_thread(void *arg) {
         page_storage.is_auto_sd_repair_active = false;
     }
     page_storage.is_sd_repair_complete = true;
+
+    if (complete_callback != NULL) {
+        complete_callback();
```
SumolX, line 589: The way complete_callback() is currently implemented can potentially blow up the stack for this given thread since the calling thread is not exited until all the nested callbacks have completed.

src/ui/page_wifi.c
```diff
@@ -316,6 +316,10 @@ static void page_wifi_update_settings() {
             system_script(WIFI_STA_ON);
         }
     }
+
+    if (complete_callback != NULL) {
+        complete_callback();
```
SumolX, line 321: Same issue here as well.... what happens if this is invoked by the main thread and we keep executing nested complete_callback functions.... this could potentially cause the main thread to die and could make adding new deferred functions difficult because we will not know why initially valid new functionality is crashing because the stack memory or frames have been exhausted.

SumolX:
Overall I really like these changes.... however instead of using a function pointer (callback_complete) why not pass integer by address or boolean and set it to true.  *post_bootup_action_done = true;

src/ui/ui_main_menu.c
```diff
@@ -349,39 +354,33 @@ void main_menu_init(void) {
...
-                        post_bootup_actions[i]->post_bootup_run_complete = NULL;
-                        post_bootup_actions[i] = NULL;
-                        ++last_bootup_action;
-                    }
-                }
-            }
-        }
+    if (!bootup_actions_fired) {
+        bootup_actions_fired = true;
+        handle_bootup_action();
```
SumolX, line 383: Here you can pass bootup_actions_fired to handle_bootup_action(&bootup_actions_fired) and continue to pass the boolean all the way to 

(*next_bootup_action++)->post_bootup_run_function(bootup_actions_fired)

src/ui/page_storage.c
```diff
@@ -567,13 +567,13 @@ page_pack_t pp_storage = {
     .on_right_button = page_storage_on_right_button,
     .post_bootup_run_priority = 50,
     .post_bootup_run_function = page_storage_init_auto_sd_repair,
-    .post_bootup_run_complete = page_storage_is_sd_repair_complete,
```
SumolX, line 570: This function needs to be called otherwise it may break the input_thread which monitors the SD Card mounting if I'm not mistaken.

Master92: You're absolutely right and I actually missed that this method has side effects. I'll fix that in a minute!

SumolX:
> @SumolX I just had an idea: how about I add a new method for handling the post bootup actions for each the wifi and storage pages that in turn only execute the existing methods plus an optional callback? That way, I get to finish my callback approach and everyone else keeps their existing method signatures without unnecessary `NULL` arguments!

That sounds great!

SumolX:
Looks great but I haven't had time to test this.  Perhaps later tonight I can try it out otherwise let's get these modifications merged.

PR #459, opened 2024-11-18
Prevent possible invalid memory accesses by sprintf

Master92:
## Update (2024-11-19)

Per the discussion below, I replaced all calls to `sprintf` with respective ones to `snprintf` that only writes as many chars as fit the target.

## Original description

I compile the emulator locally with gcc-14 and it outputs warnings for some of the buffers used to hold numerical values (see example screenshot below).

These changes should not do any harm but use more of the stack memory while calling the changed methods, but it should be ok imho.

grafik

### DIFF
--- a/src/core/dvr.c
+++ b/src/core/dvr.c
@@ -44,14 +44,14 @@ void dvr_update_status() {
 void dvr_enable_line_out(bool enable) {
     char buf[128];
     if (enable) {
-        sprintf(buf, "%s out_on", AUDIO_SEL_SH);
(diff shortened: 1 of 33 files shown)


SumolX:
I would also update the related code to use snprintf to prevent buffer overflows.  sprintf should not be used in any code base.

Master92:
I agree about using `snprintf` over `sprintf`. I'm just not sure if I (or you 😉) should better open another PR for that. I like being as specific about only one topic per PR as possible.

ligenxxxx:
I use `sprintf` in many places. I have indeed had an app bounce during development because the buf was not big enough.

PR #463, opened 2024-12-03
Move localization into an ini file

Master92:
As discussed in #452, I went ahead and moved the Simplified Chinese localization to an external ini-file.

This change opens the door for future additional localizations. All that has to be done is adding the localization ini-file and a single line in language.c, declaring the new language.
Of course, a configuration combo box has to be re-introduced as well, but that's only a one-time change.

I tested memory usage in the emulator and it seems fine. The results were acquired with:

```c
struct mallinfo mi = mallinfo();
LOGD("Total heap size: %d bytes", mi.arena);
LOGD("Free heap space: %d bytes", mi.fordblks);
```

gathering the following results:

```
Previous:
Total heap size: 135168 bytes
Free heap space: 133792 bytes

New:
Total heap size: 135168 bytes
Free heap space: 103120 bytes
```


ligenxxxx:
image
image
`Firmware   ` cannot be parsed correctly, maybe it has something to do with the space at the end of the string?

ligenxxxx:
I also found that ini does not support \n. I can use character concatenation to solve it.

PR #466, opened 2024-12-10
enable TP2825 color killer

ligenxxxx:
This fixes the issue where the analog module displays other colors in the menu (which is only black and white).


nerdCopter:
Commit `c224c17`produces pink video for a second or two before the analog signal syncs (rapidfire).  Does this PR also fix such? (i did not test yet)

sanderpuh:
Just tested the build, here's my findings so far: https://www.youtube.com/watch?v=y1qEbA5UOf0

Have not been able to link the behavior to specific actions, but it is very unstable. Using PAL camera and PAL setting in the goggles. Navigating for a while does return a decent image in the end.

- It seems to switch to NTSC for the some reason at some point navigating the module
- The menu is unreadable at times
- Pink overlay is showing on the image sides after returning to the camera view

PR #470, opened 2024-12-28
Support for Russian language

Sat1l:
Added support for Russian language + a script that would move generated font files to the lvgl folder just because.


Sat1l:
as of now it just bootloops the goggle if Russian file is specified, but english/chinese are OK. Also if I replace characters in en/ch with cyrillic letters it just works, but with rus - nope

Sat1l:
got the goggles to boot up with a new lang but some strings do not get replaced / get torn apart / do not respect a newline symbol and just show it

ligenxxxx:
https://github.com/hd-zero/hdzero-goggle/pull/474
Please see this, I removed the `\n` in the .ini file.
Also the `firmware   ` still cannot be translated correctly.

Sat1l:
Funnily enough - the same issue as with my previous attempt at i18n happened - a **segfault happens when exiting the clock page**. This is also present in Chinese translation of the upstream.

The fixes were very nice, but i still face some issues when translating:
Sometimes there are two dots at the end of the notes, even when only one dot is placed in .ini file, this also happens in Chinese translation - e.g. in the "Recording" page both rows of the note have double dots at the end.

Sometimes the second row of the note is being chopped off randomly - in “Image” at “Make sure a HDZero VTX or analog VTX is powered on for live video” my translation gets cut off at “… для просмот”.
In the “recording” tab its at “TS format is highly recommended” and the translations after “Рекоме” is being cut, very early in the string.
Input menu gets a cut at “’Toggle source' will switch between HDZero and Expansion module”, after “… перекл”

ligenxxxx:
https://github.com/hd-zero/hdzero-goggle/pull/478
I solved the problem of `Clock` -> `Back` causing app crash.

ligenxxxx:
@Sat1l You will need to do some debugging to determine where the problem occurs.

Sat1l:
I was able to get everything working. Also fixed a couple key-translation mismatches between UI and translations. 

There was no Chinese translation for "Element" (can be seen in ui_osd_element_pos), so I added the needed row with "tbd" to be fitted with appropriate translation

Sat1l:
Also in [page_wifi.c](https://github.com/hd-zero/hdzero-goggle/blob/main/src/ui/page_wifi.c) "scroll to cancel" is not internationalized - I could give it a crack but looks like it would take me much more time then it will you.

Sat1l:
The translation is now complete.

pitts-mo:
@Sat1l can you please verify wording changes for the source page and provide feedback?

Signal detected = "Сигнал обнаружен"
No signal = "Нет сигнала"
Analog input requires Expansion Module = "Для аналогового входа требуется модуль расширения"

Snapshot_2025-06-06_15-18-28

PR #473, opened 2025-01-21
analog video support 16:9

ligenxxxx:
You can set analog video as 16:9 or 4:3 in the Source menu page


cynfewl:
Thank you so much for adding this feature!

I just finished flying some analogue quads after updating and 16:9 with a 4:3 camera is so much fun!

PR #479, opened 2025-02-06
Automatic recording HDMI in

ligenxxxx:
When goggle displays HDMI in, if the SD card exists and has more than 100MB of remaining storage, it will automatically record.
- h264 encoding
- 34 mbps
- TS format only
  
I tested:
- 1080p30
- 1080p50
- 1080p60
- 720p50
- 720p60
- 720p90
- 720p100*

Note that HDMI in 720p100 will be recorded at 720p90.

### DIFF
--- a/src/core/app_state.c
+++ b/src/core/app_state.c
@@ -97,7 +97,9 @@ void app_switch_to_hdmi_in() {
     IT66121_close();
     sleep(2);
 
-    if (g_hw_stat.hdmiin_vtmg == 1)
+    if (g_hw_stat.hdmiin_vtmg == HDMIIN_VTMG_1080P60 ||
+        g_hw_stat.hdmiin_vtmg == HDMIIN_VTMG_1080P50 ||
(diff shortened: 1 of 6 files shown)


ligenxxxx:
1080p60 recording is currently unavailable.

ligenxxxx:
The issue of 1080p60 not being able to record has been temporarily resolved. I think HDMI recording is now fine.

nerdCopter:
is hundredth and thousandth decimal values allowed? i have seen `59.94` and other variations in video-files. i am unsure of what the actual standard is.

mikecarr:
Should there be an option to enable/disable this? What happens if there is space but while you are flying the disk gets full?

mikecarr:
Are there other reason why it would not record? I was able to get it to work at home but I just got back from flying and nothing was recorded.
[2025-02-20.log](https://github.com/user-attachments/files/18896639/2025-02-20.log)

PR #487, opened 2025-02-13
More sensitive motion detection

ligenxxxx:
One user reported that his Goggle went into sleep mode during the game. His head moved slightly so Goggle did not detect the movement.

### DIFF
--- a/src/core/ht.h
+++ b/src/core/ht.h
@@ -10,7 +10,7 @@ extern "C" {
 #define CALIBRATION_BCNT   8 // calibartion loop cnt = (1<<CALIBRATION_BCNT)
 #define gyroWeightTiltRoll 0.98
 
-#define MOTION_GYRO_THR    3000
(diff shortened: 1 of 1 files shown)


pitts-mo:
I question if it would be desirable to add a sensitivity adjustment for activity detection.

I have been annoyed by screen dimming and turning off if I fail to recognize this. Fortunately, this only seems to happen in sim and not in live flight.

ligenxxxx:
> I question if it would be desirable to add a sensitivity adjustment for activity detection.
> 
> I have been annoyed by screen dimming and turning off if I fail to recognize this. Fortunately, this only seems to happen in sim and not in live flight.

This really needs to be discussed. I can't decide for now.

nerdCopter:
user-configurable makes sense; However, my head is never static so it never affected me.  default 1000 :man_shrugging:

PR #503, opened 2025-03-24
utils: add script ssh-deploy.sh

DavHau:
This script uses ssh to copy a custom HDZGOGGLE binary to the goggles and replace the running process with it.

This allows for rapid testing of changes, as one doesn't have to put the file on a sdcard each time.

example usage:
```
Usage: ./utilities/ssh-deploy.sh <host> <HDZGOGGLE_binary>
```

It stops the running HDZGOGGLE process, uploads the new binary to /tmp and launches it.

The goggle fs remains unmodified and after a reboot everything is back to normal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,11 +37,14 @@ Where x.x.x is the OTA_VER.RX_VER.VA_VER
 
 ## Loading the Firmware
 
-Firmware can be either flashed via goggle menu or alternatively be executed via the SD Card with a custom development script.  An example of this development script is provided below.  The goggles automatically checks to see if the develop.sh script exists in the root of the SD Card and if found develop.sh is then executed.
+There following methods can be used to load the firmware:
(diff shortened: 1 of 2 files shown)


SumolX:
don't forget that you can utilize `/root/.ssh/authorized_keys` so you don't need to type in your password each time.

RaceMiata:
Great util!

I run into an issue with WiFi connection becoming stale so I can't ssh into it (including running the sh script again).

I have a fix in `page_wifi.c:page_wifi_post_bootup_action()` to force
 `page_wifi_bootup_pending`

flag to be false if
 `page_wifi_get_real_address()`

returns non-null.
That is, if there's already WiFi connection, then force the flag to false to force restarting WiFi app.

I also change /dev/null to /tmp/HDZGOGGLE.log in case I want log for HDZGOGGLE

PR #505, opened 2025-03-24
add declarative firmware build via nix

DavHau:
This allows developers with a linux machine or docker or wsl to build
the app via this command.

```nix
nix build .#goggle-app
```

Except the nix package manager, no other tools need to be installed and
no environment needs to be set up.

The nix build system is quite flexible and has a rich ecosystem.
In the future, the nix declarations could be extended in order to:

- add automated testing to the build pipeline
- switch to a recent toolchain instead of the outdated one from 2018
currently pulled in by `setup.sh`
- cleanup the repo from binary files like busybox, tinycurl, dropbear,
etc. Instead declare dependencies through its source and build
definition, which simplifies updating, modifying and adding stuff.
- update the currently used 4.9.X linux kernel which is end-of-life
since 2023 and likely insecure
- declare a reproducible build for the base os, so it can be more easily
extended by the community
- improve the firmware architecture and update mechanism to make it easy
to deliver updates including the base os

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,7 @@ media.tar.gz
 /.vscode/settings.json
 .DS_Store
 utilities/font/out
+
+# nix related
+result
+result-*
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Note: decompressing the repository in Windows system may damage some files and p
 ### Devcontainer Setup
 
 This repository supports the [vscode devcontainer](https://code.visualstudio.com/docs/devcontainers/containers) integration.
-To get started install docker, vscode and the decontainer extension.
+To get started, install docker, vscode and the devcontainer extension.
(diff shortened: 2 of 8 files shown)


DavHau:
> I don't understand the need for this. A function and/or alias to `bash setup.sh; make -C build;` works well enough.

The main point is to integrate this project with a proper build system.

When I look at the repo, I see binary executables, `.so` files, `.a` files, not knowing which source code they are built from. I see copies of source code of other open source projects, not knowing which version those are, or if they have been patched or not, outdated compilers being used, ...

The whole firmware consists of many components, not just the goggles app. But updating anything substantial apart from that app is hard as due to how things are organized right now.

A build system can help getting that situation under control and I'm happy to help with that and do the heavy lifting, but I need to start somewhere.

Issue #500, opened 2025-03-19
[feature request] Offload DVR from goggle via Wi-Fi


Sat1l:
It would be really nice if where was some functionality that allows to offload DVR from goggles to a NAS without the need to extract the SD card from the goggles. I also see a workflow like this:
1. You get home from a training session/competition
2. Power and leave the goggles at the bench/discharging a pack/etc
3. Goggles offload media to a NAS
4. After offloading is done and nothing is lost - wipe the goggles DVR folder

DavHau:
I was thinking about exactly this use case as well.
Basically the only thing that you would need on the goggles is an ssh server.
This would allow you to run a service on your nas that pulls the files from your goggles, whenever they appear on the wifi.

In theory an ssh server would be simple to add and this has been [attempted before](https://github.com/hd-zero/hdzero-goggle/pull/221), but the effort is stuck as it would require updating the base rootfs.

Unforunately the community cannot help with that as building the base rootfs requires a [non-open/proprietary toolchain](https://github.com/hd-zero/hdzero-goggle/issues/498) that only the hdzero devs posess.

Sat1l:
Wait, don't the goggles already have an ssh server? While digging through the settings I've seen something about ssh.

> Basically the only thing that you would need on the goggles is an ssh server.

DavHau:
Oh nice, seems like ssh is already there.
For everyone who is looking for it, it is a bit hidden. Look under: `WiFi module` -> `Page` -> `System`

PR #507, opened 2025-03-28
Media server for downloading DVR off of goggles

Sat1l:
This PR adds a simplistic DVR-preview page for offloading DVRs off of goggles without the need of extracting the SD card. (targeting issue #500)

It is somewhat slow (was able to get only 1.5MB/s consistently on 2.4ghz, oddly same speed for 5ghz), so for now the buttons for downloading all of the DVRs act as placeholders since even an hour of flying footage would be loading for hours, so speeds is an issue to solve.

The page is accessible at http://\*goggles_IP\*:8080/index.html

As of now the server launches when WiFi connects / shares in AP mode. I plan on adding the setting for enabling/disabling later.

I think this idea can be later leveraged for manipulating the goggles settings through the same webpage, maybe dumping / loading the settings.ini file would be the next thing for me when I get DVR side done.

Screenshot 2025-03-28 at 13 51 04


Sat1l:
Loading speed vary a lot, sometimes cause of the unknown conditions - I was able to get 2.5MB/s yesterday. Also my module has no antenna, maybe thats something I can improve on - I'll try later when the antenna I ordered ships to me

SumolX:
You should check out the web interface @bjgarcia and I have been working on.  We handle live video streaming and the ability to stream a DVR recording and perform file management with login authentication.  What prevented us from creating a PR was the fact the wifi was so slow.... but I've just submitted a PR that doubles the wifi speed when in Host/AP mode.

WebUI:
https://github.com/SumolX/hdzero-goggle-services/tree/webui

PR to double wifi speed:
https://github.com/hd-zero/hdzero-goggle/pull/508

Sat1l:
@SumolX woah, that's huge! Looks like a huge work has been done already, but if there's anything I could help with it would be nice, where can I reach you?

Also the fact that boosting up WiFi speeds in host mode took two rows is just crazy. Great finding. I was trying to reach the boost in client but as far as I was able to find it already works in 802.11n. My thinking was that the lack of antenna was the limiting factor, but I haven't received one yet or had the ability to acquire

SumolX:
> @SumolX woah, that's huge! Looks like a huge work has been done already, but if there's anything I could help with it would be nice, where can I reach you?
> 
> Also the fact that boosting up WiFi speeds in host mode took two rows is just crazy. Great finding. I was trying to reach the boost in client but as far as I was able to find it already works in 802.11n. My thinking was that the lack of antenna was the limiting factor, but I haven't received one yet or had the ability to acquire

On the HDZero discord server I go by `SX` (sumolx)

pluggedinn:
I was thinking of building this and I’m happy to see it being developed already. @SumolX Do you have a working MR that includes the web interface and the line changes to boost WiFi?

PR #508, opened 2025-04-04
Enable 802.11n support for double the bandwidth when running as host.

SumolX:
Title says it all.... we can now achieve close to 5.5MB/s when goggle is running as host.

Top terminal is the goggles running iperf3 as a server.
Bottom terminal is the macbook running iperf3 as a client.

Screenshot 2025-04-03 at 10 41 24 PM


SumolX:
Testing complete... Unable to get chipset working in 40Mhz mode we are receiving a performance boost with this two line change.

PR #509, opened 2025-04-12
Fix post bootup action

SumolX:
Fixed a few issues where the refactor of post bootup actions caused breakage.

1. SD Card repair thread needs to complete prior to invoking next boot action, thus we need to hold onto the bootup_complete_cb to invoke when thread is finished.

2. Sorting is now working as we need to apply indirection when passing pointers as arguments.

3. Preventing SD Card repair from running twice at bootup.  People should notice the SD Card becoming available at half the previous amount of time.

### DIFF
--- a/src/core/common.c
+++ b/src/core/common.c
@@ -14,7 +14,8 @@
 // globals
 pthread_mutex_t lvgl_mutex;
 atomic_int g_key = 0;
-atomic_int g_init_done = 0; // 0= init not done, 1= done, -1= dial/up/down pressed
+atomic_int g_init_done = 0;       // 0= init not done, 1= done, -1= dial/up/down pressed
+atomic_int g_bootup_sdcard_state; // 0= no card, 1= has card, -1= finalized
 static hw_revision_t g_hw_revsion = HW_REV_UNKNOWN;
 ///////////////////////////////////////////////////////////////////////////////
 
--- a/src/core/common.hh
+++ b/src/core/common.hh
@@ -34,6 +34,7 @@ hw_revision_t getHwRevision();
 
 extern atomic_int g_key;
 extern atomic_int g_init_done;
+extern atomic_int g_bootup_sdcard_state;
 extern pthread_mutex_t lvgl_mutex;
 
 #ifdef __cplusplus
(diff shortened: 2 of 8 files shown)


SumolX:
Need a better approach to determi when to actually mark the sd card repair is completed in order to allow other post boot up operations to continue once the sd card is mounted.  Marked PR as draft for now.

SumolX:
Bootup Tests:

1. No SD Card... ensure Wifi bootup action is invoked
2. SD Card found... ensure Integrity Check completes and is only executed once then verified Wifi bootup action was invoked after.


Running Tests:
1. Removed SD Card and re-inerted card.... verified Integrity Check was invoked.

Overall we fixed many issues here.  Prevent double Integrity checks from occurring, ensure priority was enforced, ensured actions are completed properly before kicking off the next action and finally introduced logic to properly handle SD Card presence when booting and during runtime.

nerdCopter:
Flashed a mashup of this plus PR 495.

Booting with SDCard: the initial menu shows "SD Card", then when the feed screen shows (HD or Analog) the SD Card will be marked with red-question for a few (2 or 3?) seconds then sdcard icon becomes good.

Booting without SDCard: The initial menu shows "SD Card" even though it is out. Then normal behavior.

When the goggles display a feed, the sdcard insert/removal icon is rather quick.

When the goggles are not locked into an analog feed (black instead of white/black noise), then the sdcard icon may not show for a few seconds (3 or 4?).  I beive this is nothing to do this this PR per se. i think it is somthing about the non-locked analog state and the OSD output.

Overall SDCard seems faster (placebo?).

SumolX:
> Overall SDCard seems faster (placebo?).

I'll take that...lol. but it should be faster because we aren't running an Integrity Check twice.  Overall it sounds like you had no new issues introduced with this PR which is good news.

The only new behavior if you boot up into menu instead is that you will see "WiFi: Pending" now as this post bootup action is waiting for the Integrity Check to complete per design.  Then once the Integrity Check does complete..... then "WiFi: Searching" should be reported if running in WiFi Client mode or the SSID you have chosen for WiFi Host mode.

PR #495, opened 2025-03-07
Prevent update from bricking goggle

pitts-mo:
This PR changes both the update_goggle.sh and write_flashes.sh scripts to prevent writing a bin into an /dev/mtd8-10 of unexpected size.

Notes:
-prevents soft bricking of goggle VA or RXs
-needs formal testing on fully working goggles (neither of the RX in my goggle are currently mounting to /dev/mtd8/9)
-beeps a quick double chirp on success finding the expected size of each /dev/mtd* 
-beeps two long and a chirp on failure to find the expected size /dev/mtd*
-does not automatically recover the google however commented lines in write_flashes can allow for easy recovery of erroneously flashed VA and RX. 

To recover a broken VA or RX place these files on SD card to :
HDZGOGGLE_VA.bin  # from Recovery if  VA is failing (no video on goggle)
HDZGOGGLE_RX.bin # from Recovery if one or more RX are failing (two or four HDZero signal indicators stuck)
develop.sh # copy of write_flashes.sh and uncomment  check_mtd_write for HDZGOGGLE_RX/HDZGOGGLE_VA as needed

Other SD card files that may help while trying to recover:
self_test.txt # empty file to dump diagnostic logging into
DONOTREMOVE.txt # empty file -write_flashes will leave .bin files on SD card. However, after goggle display returns remove the HDZGOGGLE_VA.bin file from SD card. When both HDZero RXs have been verified remove the HDZGOGGLE_RX.bin from SD card.

PR #513, opened 2025-05-02
firmware update / state indication improvement

pitts-mo:
firmware update / state indicator improvements:
-prevent soft bricking VA / RX (very rare)
-add start, success and failure beeps to existing end beep indicator of recovery flash
-add beep indication for failure during gui flash
-silenced beep indications for successful gui flash
-erase entire mtd before write
-fix gui status bar indication
-simplified scripts to improve consistency
-display distinct RX versions (indicate RX firmware issue/mismatch)

------
You *SHOULD* never see a mismatch in RX versions. However, It did happen to me at least once.

I had considered displaying both RX versions only when a mismatch is detected. However, it is more accurate and appears to promote user awareness to persistently report the distinct RX versions.


pitts-mo:
tested:
-install from gui
-reinstall from gui
-boot with VA/RX bins and DONOTREMOVE.txt updates VA/RX and leaves VA/RX files on SD card
-boot with VA/RX bins without DONOTREMOVE.txt updates VA/RX and removes VA/RX files from SD card

pitts-mo:
notable change: now erase entire mtd before write (prior code only erased initial 65536 of rx mtd)

pitts-mo:
This resolves failure of gui status bar during mtd flash and disables successful mtd flash beeps during normal gui update introduced by PR #495.

I think this is now good enough to consider for merge.

Potential items for future enhancement:
-detailed failure indications in gui?
-automatic failure logging to SD?

PR #525, opened 2025-06-05
Merged BoxPro and Goggle Codebases

SumolX:
In order to lessen the effort for maintaining two separate product lines we have consolidated both the BoxPro and Goggles into the hdzero-goggle codebase.

Readme has been updated and includes instructions on how to build both targets.  setup.sh will automatically generate to build directories:

build_goggle
build_boxpro

Cleanup will be performed as a phased approach but in its current form I tried to make maintainability a priority for both product lines.  The use of the preprocessor has been minimized as much as possible and we can leverage `getTargetType()` to identify which products within the codebase.

Please test this codebase as much as possible but I do believe I have ironed out all merge issues initially encountered.

### DIFF
--- a/src/core/osd.c
+++ b/src/core/osd.c
@@ -370,27 +370,28 @@ void osd_channel_show(bool bShow) {
         snprintf(buf, sizeof(buf), "  To %s?  ", channel2str(g_source_info.source == SOURCE_HDZERO, g_setting.source.hdzero_band, ch));
         lv_obj_set_style_bg_opa(g_osd_hdzero.channel[is_fhd], LV_OPA_100, 0);
     } else {
-        if (g_source_info.source == SOURCE_HDZERO)
+        if (g_source_info.source == SOURCE_HDZERO) {
             ch = g_setting.scan.channel & 0x7F;
-        else if (TARGET_BOXPRO == getTargetType() && g_source_info.source == SOURCE_AV_MODULE)
+        } else if (TARGET_BOXPRO == getTargetType() && g_source_info.source == SOURCE_AV_MODULE) {
             ch = g_setting.source.analog_channel & 0x7F;
-        else
-            return;
+        } else {
+            bShow = false;
+        }
 
-        color = lv_color_make(0xFF, 0xFF, 0xFF);
-        snprintf(buf, sizeof(buf), "CH:%s", channel2str(g_source_info.source == SOURCE_HDZERO, g_setting.source.hdzero_band, ch));
-        lv_obj_set_style_bg_opa(g_osd_hdzero.channel[is_fhd], 0, 0);
+        if (bShow) {
+            color = lv_color_make(0xFF, 0xFF, 0xFF);
+            snprintf(buf, sizeof(buf), "CH:%s", channel2str(g_source_info.source == SOURCE_HDZERO, g_setting.source.hdzero_band, ch));
+            lv_obj_set_style_bg_opa(g_osd_hdzero.channel[is_fhd], 0, 0);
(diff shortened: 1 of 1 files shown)


pitts-mo:
Oh so Magical, thank you. -p

pitts-mo:
OSD element "CH:--" persists when goggle is booted directly to AV in.  Switching to another source then back to AV in will remove the CH:-- OSD element. I did not test this on prior releases but suspect this may be related to boxpro code merge.

I do not find any change that should block merge. But suggest we release note specifically calls out source page changes.

Some batch 2 HDZero Goggle items used to evaluate general user experience:
-goggle update from gui
-re-goggle update from gui (test update_goggle.sh and display of firmware version string)
-RX/AV recovery flash (test write_flashes.sh)
-HDZero source
-"analog" source (module w/ v2 goggle power control)
-HDMI in source 
-AV-in source
-image brightness adjust ( from roller input )
-OSD -elements toggle DATE TIME
-power -cell average and analog RX Auto
-record w/manual, .ts, date
-ELRS -DVR start/stop and HDZero channel change
-WiFi AP -rtsp and ssh
-playback -browse and play ( desire ability to toggle/adjust small file size filter -some day :-) )
-focus chart
-clock -24 hour
-input -roller brightness
-Go Sleep -enter/exit while analog module is enabled (w/ v2 goggle analog bay auto power)

pitts-mo:
Confirmed this does not occur on prior goggle releases.

OSD element "CH:--" persists when goggle is booted directly to AV in (auto scan -last, last w/AV-in).  Switching to another source then back to AV in will remove the CH:-- OSD element.

SumolX:
However, what this does not fix is the root cause of the issue.... where using last AV_IN as source... we should not be hitting this logic.  Which tells me that between main::start_running and osd::osd_channel_show we have a bug somewhere that existed in the boxpro codebase.

I would enter a new defect after we submit this PR as @pitts-mo had suggested.

nerdCopter:
did not "deeply" test, did notice `771510e` did not go to my "last" input (analog).

pitts-mo:
Sorry, I have just tried 771510e from SD card. 

It now persists the CH:-- element in analog / AV-in and I can no longer clear it by switching sources. 

I have not had issue booting my goggle to last source with the recent builds in this PR when I have Auto scan set  to Last, Last.

nerdCopter:
> That's strange.... the change made would have only prevented the analog RF channel from being displayed on goggles. I did test last analog on both boxpro and goggles and it was working. I will retest and see if it's an issue or not.... but I think @pitts-mo also reported that working just fine.

it may (probably) be a `main` mainline code.  i have seen it before, but random. only tested this PR two bootups.

SumolX:
> did not "deeply" test, did notice `771510e` did not automatically go to my "last" input (analog). (OG goggles Nov-2023 order)
> 
> EDIT: i did not test the prior commit. i was able to get HD vs Analog via: image and did not consider trying the menu.

Thanks for the reference!  I will look at it tonight!

pitts-mo:
CH:-- is annoying as since https://github.com/hd-zero/hdzero-goggle/commit/771510ee573094bed4026af6deac8b0585dc7d06 it is now showing on boot of analog and you need to switch to HDZero source and back. to clear it (or disable CH: OSD element all together?)

Regardless this PR is a required step forward and so far I have not found/confirmed a bug serious enough to prevent merge.

nerdCopter:
PEBKAC -- i only had scan set to last, not input. sorry for the added "noise" and waste of your time!

PR #538, opened 2025-06-22
Update setup.sh -add toolchain awareness

pitts-mo:
updates to setup.sh script will now:
-prompt the source code user before downloading tool chain (~50mb)
-add verbosity for download and extraction status
-preserve downloaded tool chain .tar.bz2 file for future use
-allow user to pass full path of tool chain file or manually copy tool chain file into root of source to prevent prompting or new download of tool chain

These changes potentially reduce network data usage, increase tool chain portability during toolchains.bootlin.com service interruptions (~2025-06-21), and promote general awareness of source code users.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,12 +1,31 @@
 #!/bin/bash
-set -e
+echo "Working directory: `pwd`"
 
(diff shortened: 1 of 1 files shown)


pitts-mo:
I was afraid of this.... closing to stop github build runaway

PR #553, opened 2025-07-18
merge goggle2

ligenxxxx:


ligenxxxx:
@pitts-mo Could you please modify update_goggle.sh?

pitts-mo:
@ligenxxxx I see this PR reverts the unified check before write routine and now breaks out distinct calls for : 
update_boxpro.sh
update_goggle.sh
update_goggle2.sh

Are you suggesting all three of these distinct scripts be updated to support the unified check before write routine?

pitts-mo:
I did like how @SumolX merged goggle and boxpro update_goggle.sh but also recognize it might feel safer to manage updating different hardware with distinct scripts.

SumolX:
@pitts-mo and @ligenxxxx the merged update script was possible because both the box pro and goggle v1 were identical in hardware access calls and the only difference between either were file names.  If that's not the case for googles v2..... then maybe keep them separate but at the same time we can do target checks as well which makes maintenance a little easier across the entire codebase.

Personally I would like to keep a single file.

ligenxxxx:
> I see update similarities far outweigh the differences. Merging all three update_goggle scripts would be better for consistency and code maintenance. I will still investigate keeping platform specific launcher scripts to provide support distinction.
> 
> questions:
> 
> 1. What functionality does the second copy of VA firmware written at 8388608 of goggle2 add? (independent display process?)
> 2. Does goggle 2 need the second VA segment to be written when being recovered by write_flashes.sh?
> 3. What method can be used to read the version string of goggle2's second VA instance ? (sys_ver->va = I2C_Read(ADDR_FPGA, 0xff);?)
> 4. Is there any reason for goggle 2 VA's being updated independently? / or can the entire VA flash be erased and written each time?
> 
> Note: I plan to update the update_vtx.sh script. Without ability for drivers to identify specific MTD devices it's not easy to prevent soft bricking an RX but the size check can at least prevent VTX update from soft bricking the VA. eg: https://discord.com/channels/880782999006883870/880887864357953677/1398059604931383427

Goggle 2 writes the FPGA firmware to two places because occasionally the FPGA fails to load the firmware when it is powered on, for unknown reasons.
However, it succeeds when the FPGA loads the second firmware.

PR #554, opened 2025-07-29
merged update scripts to include goggle2

pitts-mo:
-work around goggle2 firmware file names matching goggle v1 firmware file names
-add secondary VA write for goggle2

page_version.c should be able to  run update_goggle.sh to update goggle, boxpro, and goggle2.

Notes:
-these changes are completely untested! (family tasks)
-extended from @SumolX boxpro merge. (seems like the best way to handle unified update scripts)
-migrate goggle 2 firmware files to distinct names?
-update_vtx.sh not changed (already appears to perform a destination flash size check.)

### DIFF
--- a/mkapp/app/script/update_goggle.sh
+++ b/mkapp/app/script/update_goggle.sh
@@ -1,24 +1,29 @@
 #!/bin/sh
 
 PLATFORM="$(cat /mnt/app/platform)"
+PLATFORMfile=$PLATFORM
+if [ "$PLATFORM" == "HDZGOGGLE2" ];then
+  # work around goggle2 firmware file names matching goggle v1 firmware file names
+  PLATFORMfile=HDZGOGGLE
+fi
 TMP_DIR=/tmp/goggle_update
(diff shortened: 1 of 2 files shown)


pitts-mo:
I only have a google v1. 
I have tested these scripts on HDZGOGGLE and simulated HDZGOGGLE2 and HDZBOXPRO to work as expected.

Suggest independent install and reinstall tests for all three platforms.
-p

pitts-mo:
Please check PR #554 

Then page_version.c  can have this lines:
#if defined(HDZGOGGLE)
    char shell_path[] = "/mnt/app/script/update_goggle.sh";
#elif defined(HDZBOXPRO)
    char shell_path[] = "/mnt/app/script/update_boxpro.sh";
#elif defined(HDZGOGGLE2)
    char shell_path[] = "/mnt/app/script/update_goggle2.sh";
#endif

replaced with:
    char shell_path[] = "/mnt/app/script/update_goggle.sh";

PR #560, opened 2025-08-23
Restore external AnalogRX power function

pitts-mo:
-Remove conditions for initial setting of AnalogRX power.
-Remove AnalogRX power state change beep.
-Restored AnalogRX power state checking to source selection.

The AnalogRX power state change beep was desirable for testing but seemed to just prompt desire for an optional goggle ready beep feature.

### DIFF
--- a/src/driver/dm5680.c
+++ b/src/driver/dm5680.c
@@ -325,17 +325,9 @@ void DM5680_ResetHDMI_RX(uint8_t on) // Reset HDMI_RX,0=reset
 
 void DM5680_ExternalAnalog_Power(uint8_t on) {
     uint8_t Cmd[5] = {0xAA, 0x55, 0x02, 0x6, 0x00};
-	#if defined(HDZGOGGLE)
-      // Note Cmd[4] = 0 means power on for external analog module
+	      // Note Cmd[4] = 0 means power on for external analog module
 	Cmd[4] = !on;
-    #elif defined(HDZBOXPRO)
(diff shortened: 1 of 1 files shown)


ligenxxxx:
I tested on BoxPro:
- From the source menu, go to analog/an in/hdzero, and the analog module is powered correctly.
- Using the `toggle source`, the analog module is powered correctly.

pitts-mo:
If goggle2 is verified I plan to add additional logic for goggle2 internal/external selection.
if analog_module = 1 expansion module is turned ON else (analog_module = 0) expansion module will turn it OFF.

ligenxxxx:
I noticed that when merging the Google2 code, I introduced some errors in the analog section, causing the external analog module to malfunction.

pitts-mo:
This PR has been good for me on Goggle v1. If external analog RX power works as expected for v2 it should be good to merge. We can always circle back to add v2 logic enhancement for internal vs external as desired.

FLORIANSV35:
@ligenxxxx Hi ! I was the one who did the test with the goggles 2.

pitts-mo:
Good then then it sounds like the base function of this is now working as expected for goggle v1 and v2.

I like the idea of merging this as-is but I am also interested in creating a separate PR to extend Analog RX Power Auto function into the users internal vs external selection.

pitts-mo:
I suspect the external analog video not working may be e separate issue... unless we need to force the internal modules power off.

PR #579, opened 2025-11-03
Use umlauts in German translation

Master92:
As a native speaker I was really confused that instead of umlauts, the German translation used two characters. Obviously, this was based on the lack of characters in the font files.

I added the missing characters to the Spanish character generation script and adjusted the translation ini-file accordingly.


Master92:
I just saw that @ligenxxxx introduced those changes in #576. However, all that would have to be done was adding the appropriate characters to the lvgl font library. Therefore I just reverted his changes on top of my adjustments and force pushed right now.

PR #594, opened 2026-01-28
Fix DVR start/stop logic

RaceMiata:
Fix a couple scenarios for the ELRS backpack DVR controls:

1. Activate a DVR-start command with 0-second delay to start DVR immediately. Any previous time-delayed DVR-stop command should have the timer discarded or else the 0-second delay command will be over-turned when the stale timer expires.

2. Invoke a delayed-start command. Before the timer expires invoke a delayed-stop sequence. DVR is not started at this point but if a manual DVR-start (from goggle button) is invoked, any former time-delayed commands should be discarded or else the manual DVR-start command will be over-turned when the stale timer expires.

I come across scenario 2 quite often. I set ELRS backpack with 5s-start timer and 15s-stop timer. The 15s-stop timer is to prevent creating separate clips on a flight when I do flip-over-crash disarm-rearm sequences. The 5s-start is to prevent initiating flight footage when I test-arm at staging.

When I plug in to stage for a race, I test-arm to make sure motors turn properly by arming for less than 5s. With the 5/15 settings, the short arming won't start DVR. I then get to position and have the option to start DVR manually before arming. If the manual DVR-start happens before the 15-s timer expires then DVR stops when the timer expires. In reality, once I start DVR manually the backpack 5/15 timers should be discarded.

Added new flags record_pending and g_sdcard_ready. They are used to restart DVR automatically for whatever happens to the SD card being not ready (e.g. error writing file, SD card eject/re-insert), and more importantly facilitate user to toggle DVR on even before the SD card is ready (e.g. on bootup, after re-insertion). Extend SD-card-? icon to not just when SD card is out but also when SD card is being repaired after mounted.

Real world use case: I have to reboot goggle for whatever reason when other pilots are ready at a race round. Judging solely by the "sd card inserted" OSD icon turning off doesn't mean the SD card is ready. I have to wait til the automated SD card repair process to finish (no indication on OSD) before I cam start DVR and tell race director I'm ready. Otherwise, if I toggle on DVR before the SD card is ready then DVR will never start; thus, wasting precious time for every pilots for me to start DVR.

With the new feature I can just toggle DVR on as soon as goggles gets to live feed screen, and tell race director ready and the DVR will start when the SD card repair process is finished.

### DIFF
--- a/src/core/osd.c
+++ b/src/core/osd.c
@@ -174,7 +174,7 @@ void osd_rec_show(bool bShow) {
         return;
     }
 
-    if (!g_sdcard_enable) {
+    if (!g_sdcard_enable || !g_sdcard_ready) {
         osd_resource_path(buf, "%s", is_fhd, noSdcard_bmp);
         lv_img_set_src(g_osd_hdzero.sd_rec[is_fhd], buf);
         lv_obj_clear_flag(g_osd_hdzero.sd_rec[is_fhd], LV_OBJ_FLAG_HIDDEN);


RaceMiata:
Corrected some logic in the last 2 commits on dvr_update_status()

PR #596, opened 2026-02-01
Fix analog ghosting

RaceMiata:
Analog input ghosting is caused by TP2825 analog processor output voltage too high. To fix or at least mitigate the issue, the AGC (register 0x25) gain can be raised to prevent clipping. TP2825 registers 0x11 and 0x12 containing brightness and contrast values are also tweaked to match resulting range for digital input.

Also the contrast range of 47 in the goggle menu does not use up the full range for the FPGA register so a conversion was made map the full range of the slider to the full range (0-255) applicable.


RaceMiata:
The previous workaround to mitigate ghosting by limiting brightness/contrast to 7/14 has now been removed. While this Pull Request significantly reduces ghosting for high brightness/contrast settings, it's still possible to have to set too high brightness/contrast from the menu to cause ghosting. Please use caution when setting brightness/contrast.

RaceMiata:
Reverted change to register 0x25 for the AGC gain. Forced AGC params in 0x21-0x27 registers for more predictable video signal range. No more delayed blow-out recovery. I tried to test weak signal situations as much as I could for possible adverse/unexpected behavior but I can't be sure I didn't mess up anything so I'm keeping this PR WIP. More testing are welcome, especially focus on weak signal, blow-out, etc.

RaceMiata:
Reverted all previous AGC-hack to try reduce peak brightness.

Implemented all-new approach by orbiting the whole analog video frame like how goggle OSD elements get orbited. It uses the existing OSD orbit setting values:

Off = no orbit,
MIN = orbiting in a square route (i.e. up, right, down, left),
MAX = orbiting in a diamond qoute (i.e. up&right, down&right, down&left, up&left).

MAX will see a bigger diagonal-pixel jump (vs 1-pixel jump in the x or y direction).

Since OSD fonts mostly have pure horizontal and vertical strokes (italic fonts are used), moving 1 pixel in only 1 of x or y direction keeps many white pixels still white. For best results in reducing analog ghosting, default Betaflight fonts are good. Avoid using big Betaflight fonts.

PR #599, opened 2026-02-11
Move merged goggle2 code in ifdef

RaceMiata:
Back when goggle2 code got merged into this unified target repo the particular line of code from hdzero-goggle2 repo was put in the common section causing AV-In not showing image on BoxPro. The code should go inside the block of if defined(HDZGOGGLE2).


RaceMiata:
Removed the I2C_Write() call altogether. It's redundant when already invoked in Source_AV() that calls rtc6715_power().

PR #604, opened 2026-03-25
Fix color problems with HDMI-In at 90Hz+ modes

RaceMiata:
Symptom: HDMI-In modes at 90Hz and up get wrong colors. It may start out looking normal but as time goes on even without any change on the screen the colors start shimmer and shift to the point they get very much off.

Having had some discussions and diagnostics on discord with "Alf Skaar", I started looking at the vtmg code and noticed with case HDMI_VTMG_720P100 it makes system call to change vdpo to 720p30 mode. I wonder why it's not set to 720p90 instead so I tried and it fixed the color problem on my BoxPro! I can set up Windows to the listed 1280x720p100Hz mode or my 1280x720p90Hz mode (with my custom timing) no color issues.

I'm not sure if the code change should apply to goggle 1 and 2 though since I haven't heard anybody having same problem with them. I don't have goggle 1 or 2 so should I make the code change for google 1 and 2 also and let people have those goggles to test?

### DIFF
--- a/src/driver/hardware-boxpro.c
+++ b/src/driver/hardware-boxpro.c
@@ -937,7 +937,7 @@ int HDMI_in_detect() {
                         break;
 
                     case HDMIIN_VTMG_720P100:
-                        system_exec("dispw -s vdpo 720p30"); // 100fps actually
+                        system_exec("dispw -s vdpo 720p90"); // 100fps actually
                         dvr_update_vi_conf(VR_540P90);
                         g_hw_stat.vdpo_tmg = VDPO_TMG_720P100;
                         I2C_Write(ADDR_FPGA, 0x8D, 0x04);


ligenxxxx:
This is not the correct way to fix it. vdpo 720p30 actually outputs 720p100.
`pclk_phase_set` You should investigate how to use this function to modify the clock timing of the IT66021. Use You can use `/mnt/extsd/pclk_phase.cfg` to modify pclk_phase value.

ZdrytchX:
> This is not the correct way to fix it. vdpo 720p30 actually outputs 720p100. `pclk_phase_set` You should investigate how to use this function to modify the clock timing of the IT66021. Use You can use `/mnt/extsd/pclk_phase.cfg` to modify pclk_phase value.

On Boxpro I get massive smearing/motion blur, and there's horrendous colour exposure loss so the image is visibly darkened except the brightest parts and there's massive contrast loss. If pixels are gettingly only 1/3rd the "exposure" needed and are updating 3 times faster than they should, that probably explains why the picture quality is incredibly poor.

Unlike RaceMiata's symptoms, the problem on the boxpro is immediately present, it doesn't just "gradually" get worse, it's instantly noticibly worse with dithering with poor contrast and gradually shifts until more than half the contrast range is just gone completely in the blacks.

PR #612, opened 2026-06-05
Support HDZero Lowband ELRS channel switching

lesharodin:


lesharodin:
sorry its was a draft

PR #613, opened 2026-07-14
Race timer sets DVR filenames to the heat you flew, e.g. hdz_0042-WinterCup-Mains-A-R5-H1.mp4

bob9:
## Summary

Lets a race timer name the next DVR recording (e.g. `WinterCup-Mains-A-R5-H1`) by sending a new `MSP_SET_DVR_NAME` (0x030F) message through the ELRS backpack.

- `elrs.c` handles the new MSP command and hands the payload to `dvr_set_race_label()`.
- `dvr.c` stores a filename-safe copy of the label (alnum kept, spaces become `-`, everything else dropped) with a 5-minute TTL so a recording made well after the heat doesn't inherit the race's name; an empty payload clears it. The label is written to the record conf on every update so a stale value never survives.
- The record process (`record/`) appends `-label` to the stock filename, just before the extension, for both naming modes — `hdz_0042-WinterCup-Mains-A-R5-H1.mp4` (contiguous) or `20260718-141502-WinterCup-Mains-A-R5-H1.mp4` (date). Snapshot `.jpg` previews and split segments share the name. Because the stock prefix and index position are untouched, the existing contiguous file counter handles labelled names with no changes.

Labels are capped at 43 chars (`REC_labelMAXLEN` 44) — the worst case (date naming + label + `.mp4`) is 63 chars, still inside the playback list's 64-char name buffer. Over-long labels are silently truncated at both entry points (MSP payload and record conf), never rejected.

Additive only — recordings without a label are named exactly as before.

## Related PRs

- ExpressLRS/Backpack#232 — companion firmware PR: the VRX backpack forwards `MSP_ELRS_BACKPACK_SET_DVR_NAME` to the goggles over the serial link, which is what delivers this message.
- ExpressLRS/Backpack#231 — syncs the backpack/goggle RTC so the named recordings also carry correct timestamps.
- ExpressLRS/Lua-Scripts#18 — the ELRS Lua script now seeds that clock automatically whenever it connects to the TX module.
- #614 — lets pilots switch between Raceband and Lowband via the backpack; part of the same integration, independent of this PR.

- ExpressLRS/ExpressLRS#3701 — TX-module side of that time sync: forwards the handset time to the backpack.

### DIFF
--- a/src/core/dvr.c
+++ b/src/core/dvr.c
@@ -1,8 +1,11 @@
 #include "dvr.h"
 
+#include <ctype.h>
 #include <pthread.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
+#include <time.h>
 #include <unistd.h>
 
 #include <log/log.h>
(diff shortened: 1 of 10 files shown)


bob9:
Pushed a follow-up commit that changes how the label lands in the filename, after trying the first cut at a race meet on paper.

Originally a label switched recordings to a separate `YYYY-MM-DD-NNNN-label` scheme. That reinvented naming the goggles already do and needed a second index scanner to keep the counter contiguous. Now the label is simply appended to whatever the stock naming produces — `hdz_0042-WinterCup-Mains-A-R5-H1.mp4`, or the same with a date-naming prefix — so a pilot's card sorts exactly as it always has, labelled and plain recordings interleaved in recording order. The extra scanner is deleted; the diff got 48 lines smaller.

Also raised the label cap from 32 to 43 chars so the race description can carry the meeting name properly. 43 is the exact budget: worst case is date naming (15) + dash + label + `.mp4` = 63, one under the playback list's 64-char name buffer. Over-long labels truncate silently at both entry points rather than being rejected.

Description updated to match.

PR #614, opened 2026-07-14
Let pilots switch between Raceband and Lowband via the ELRS backpack

bob9:
## Summary

Lets pilots' goggles be switched between **Raceband and Lowband via the ELRS backpack**. The backpack's `MSP_SET_BAND_CHAN` message carries an index into the standard 48-entry VTx table (bands A/B/E/F/R/L × 8), but the goggles currently only act on indices that map to Raceband — the eight Lowband entries (L1–L8) are silently dropped. So whatever is driving the backpack (a race timer, for example) could move the goggles around Raceband, but never onto Lowband frequencies, even though HDZero goggles tune them natively.

**Why it helps pilots:** at bigger events, organizers assign channels across both Raceband and Lowband to fit more pilots in the air. With this change the race timer can retune every pilot's goggles to their assigned frequency — including Lowband — over the backpack, so nobody fumbles through goggle menus between heats and nobody accidentally tunes over a live pilot.

## Implementation

- `hdz_index2ch()` becomes `hdz_index2bandch()`, mapping a VTx table index to an HDZero band mode + channel (L1–8 → Lowband, R1–8 → Raceband) and rejecting indices with no HDZero equivalent.
- `channel_channel_hdzero()` now takes the band into account: when it changes, it applies the same steps as the source menu's Raceband/Lowband toggle (band setting, channel label, ini persist) before re-tuning.

No behavior change for plain Raceband channel switches.

## Related PRs

Part of the same race-timer integration, though this PR is independent and can merge in any order:

- #613 — names DVR recordings after the race, pushed by the race timer over the same backpack link.
- ExpressLRS/Backpack#232 — the backpack side of that DVR-name forwarding.
- The recordings' timestamps come from the companion time-sync set: ExpressLRS/Backpack#231, ExpressLRS/Lua-Scripts#18, and ExpressLRS/ExpressLRS#3701.

This PR itself needs no backpack changes — `MSP_SET_BAND_CHAN` already carries the full 48-entry VTx table index.

### DIFF
--- a/src/core/elrs.c
+++ b/src/core/elrs.c
@@ -16,6 +16,7 @@
 #include <unistd.h>
 
 #include <log/log.h>
+#include <minIni.h>
 
 #include "core/app_state.h"
 #include "core/battery.h"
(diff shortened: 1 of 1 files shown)


liltaket:
Would love to see this merged. This bothered me last time i used lowband and switched between them with backpack...

PR #623, opened 2026-08-18
fixed HT cross-axis coupling on BoxPro (tilt was affecting pan and roll

iforce2d:
imu_orientation has one axis set as -90 + 23, presumably this accounts for the angled mounting of the daughter boards in the Goggles, but is not correct for BoxPro. The symptom was that roll and pan movements were fine, but moving in the tilt axis would cause roll and pan to be affected. I don't have either Goggles or Goggles2 to check with, but I assume the 23 degree offset is good for both.


pkendall64:
You are right @iforce2d the 23 degrees is the angle that board is mounted in the googles.

